PR #37, opened 2013-01-29
JBIDE-13440 Eclipse Application that imports projects into workspace

dazarov:
Create Eclipse Application that imports all projects from location passed as a parameter into current workspace


alexeykazakov:
Rejected as outdated.

PR #360, opened 2014-12-01
JBIDE-18837 use Base root pom instead of parent pom so we only depend on BUILD ALIAS and not the root pom version >> master

nickboldt:


fbricon:
Applied in master only

PR #364, opened 2014-12-19
JBIDE-18678 Make JBoss Usage Reporter optional

kaloyan-raev:
The org.jboss.tools.usage bundle is split into two parts API and Impl.
The API part remains with the same bundle id - org.jboss.tools.usage.
The new Impl part has bundle id org.jboss.tools.usage.impl. All parts
doing the actual usage reporting and the Usage Reporting preference page
are moved to the Impl bundle, i.e. all "internal" packages.

If the Impl bundle is missing in an Eclipse installation then the Usage
Reporting preference page is not available and no usage reporting is
executed. If the Impl is present then the usage reporting works as
usual.

A key point to achieve this is by using an OSGi service between the API
and Impl bundles. The Impl bundle declares an OSGI Declarative Service
IUsageReportingService that is consumed in the API bundle. The old
UsageReporter class is now a proxy to the UsageReportingService
implementation.

This change is backward compatible as long as the clients use only
non-internal classes from the org.jboss.tools.usage bundle.


alexeykazakov:
Replaced by https://github.com/jbosstools/jbosstools-base/pull/365 (added the new impl plugin to  feature.xml)

PR #367, opened 2015-01-13
JBIDE-18962 OpenOn for <input list=""> to corresponding <datalist id="">

dazarov:


common/plugins/org.jboss.tools.common.text.ext/src/org/jboss/tools/common/text/ext/util/Utils.java
```diff
@@ -141,6 +143,42 @@ public static String getTrimmedValue(IDocument document, Attr attr) throws BadLo
...
 	
+	/**
+	 * Returns IRegion for the value of the given attribute
+	 *  
+	 * @param document
+	 * @param attr
+	 * @return
+	 * @throws BadLocationException
+	 */
+	public static IRegion getAttributeValueRegion(IDocument document, Attr attr) throws BadLocationException {
```
alexeykazakov: Please add a unit test for this method.

dazarov: Done

PR #368, opened 2015-01-23
JBIDE-15714 Dynamic content assist for AngularJS expressions in html files

alexeykazakov:


common/plugins/org.jboss.tools.common/src/org/jboss/tools/common/text/TextProposal.java
```diff
@@ -348,7 +345,7 @@ public boolean equals(Object obj) {
      * 
      * @return
      */
-	public boolean autoActivationContentAssistantAfterApplication() {
+	public boolean isAutoActivationContentAssistantAfterApplication() {
```
maxandersen: why do we need to change this API to get angularjs content assist ?

common/plugins/org.jboss.tools.common/src/org/jboss/tools/common/text/TextProposal.java
```diff
@@ -211,9 +211,6 @@ public boolean hasImage() {
 	 * @return
 	 */
 	public int getPosition() {
-		if(position==-1 && getReplacementString()!=null) {
```
maxandersen: was this adjustment a bug ?

alexeykazakov: That was a dead code. The method was not used at all. But now we need this method. So I cleaned it up before starting using it.

common/plugins/org.jboss.tools.common/src/org/jboss/tools/common/text/TextProposal.java
alexeykazakov: This is another dead/not used API in TextProposal. But we need this method for angularjs now. There is no need to rename it, though.

PR #372, opened 2015-01-27
JBIDE-19097 performOk() and performCancel() method in TabbedPreferencesP...

alexeykazakov:
...age should return true only if all its tabs return true


alexeykazakov:
Cherry-picked to 4.2.x branch.

PR #377, opened 2015-02-19
JBIDE-19319 As-you-type validator accumulates temporary mesages

scabanovich:


common/plugins/org.jboss.tools.common.validation/src/org/jboss/tools/common/validation/TempMarkerManager.java
```diff
@@ -194,10 +195,26 @@ private IMessage addMesssage(IResource target, int lineNumber, int offset, int l
 			if(type!=null) {
 				message.setAttribute(MESSAGE_TYPE_ATTRIBUTE_NAME, type);
 			}
+			
+			if(message != null) {
```
alexeykazakov: message cannot be null here

alexeykazakov: message cannot be null here

scabanovich: Done

PR #378, opened 2015-02-20

robstryker:


dgolovin:
I see two problems:
1. It is code duplication;
2. There is no option to see license when there is no browser widget available.

For 4.2.3 (1) is OK, but we should think about sharing browser factory API through jbosstools-base in 4.3.x.
(2) should be fixed, the license fragment should show the text with link to the license at least.

PR #379, opened 2015-02-20
there needs to be a better solution here... s/4.3.0.Alpha1/4.3.0.Alpha2/g in foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/properties/internal/currentversion.properties >> master

nickboldt:


alexeykazakov:
>> there needs to be a better solution here... 
+1000!

fbricon:
Closing as out-of-date

PR #384, opened 2015-03-03
JBIDE-19272 : add JBoss Developer search engine

fbricon:


foundation/plugins/org.jboss.tools.foundation.help.ui/.classpath
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry exported="true" kind="lib" path="lib/jboss-dmr-1.2.0.Final.jar"/>
```
maxandersen: how many versions of dmr do we have in use now ? :)

i think its time we get that library into Locus instead of keep adding it to each jar.

robstryker: 1.2.0 in as.management.wildfly
1.1.6 in as.management.as7
1,1,1 in ide.eclipse.as.dmr

These three are justified in that as.dmr is exposed to clients,  while the other two are internal details to the wf / as7 management implementations.  The different versions between as7 / wf management plugins is because those are what's required for the various remoting 

It's possible that 1.2.0 would work for all cases but an inspection of the API differences between them would be required.

maxandersen: if necessary we should be able to add all three to locus. 
But at least we can avoid duplication for the basic act of parsing json.

robstryker: I think we would probably need to add 1.2.0 and 1.1.6 (or newer micro).   We definitely don't need 3. I think 2 should be enough.

fbricon:
Applied as https://github.com/jbosstools/jbosstools-base/commit/ff3e43e76fbe8c6c456f06064dbaa437af97a2bc

PR #386, opened 2015-03-05
JBIDE-19269 Batch validation problem QuickFix is not shown in the editor

dazarov:
Moving SelectRulerAction and SelectAnnotationRulerAction from org.jboss.tools.common.model.ui to org.jboss.tools.common.ui

### DIFF
--- a/common/plugins/org.jboss.tools.common/src/org/jboss/tools/common/quickfix/MarkerAnnotationInfo.java
+++ b/common/plugins/org.jboss.tools.common/src/org/jboss/tools/common/quickfix/MarkerAnnotationInfo.java
@@ -16,28 +16,31 @@ import java.util.List;
 import java.util.Map;
 
 import org.eclipse.core.resources.IMarker;
+import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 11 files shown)


common/plugins/org.jboss.tools.common/src/org/jboss/tools/common/quickfix/MarkerAnnotationInfo.java
```diff
@@ -49,101 +51,118 @@ public MarkerAnnotationInfo(List<AnnotationInfo> infos, ISourceViewer textViewer
...
 	
 	public List<ICompletionProposal> getCompletionProposals(AnnotationInfo info) {
-
-		if(info.isTop())
-			return getMarkerProposals(info);
-		else
-			return getProposals(info);
-		
+		ArrayList<String> proposalNames = new ArrayList<String>();
+		proposalNames.add("Surround with new element");
```
alexeykazakov: Can we avoid hardcoding the proposal text here? We should reference to the original WTP proposal text if possible.

alexeykazakov:
Rebased and merged.

PR #391, opened 2015-03-19
JBIDE-19474 add org.jboss.tools.foundation.help.ui to org.jboss.tools.foundation.feature/feature.xml >> master

nickboldt:


dgolovin:
Bundle is included as uncategorized  into jbosstools-base/site/category.xml.
This PR is rejected.

PR #392, opened 2015-04-03
JBIDE-19507 Search and rename participants for Batch Artifacts

dazarov:


common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/util/BeanUtil.java
```diff
@@ -74,5 +75,20 @@ public static String getPropertyName(String methodName) {
 		}
 		return null;
 	}
+	
+	public static String getDefaultBeanName(String className){
```
alexeykazakov: Please add JavaDoc to this util method.

common/tests/org.jboss.tools.common.base.test/src/org/jboss/tools/common/base/test/RenameParticipantTestUtil.java
```diff
@@ -0,0 +1,383 @@
...
+		StringBuffer sb = new StringBuffer(""); //$NON-NLS-1$
+		try {
+			byte[] b = new byte[4096];
+			while (true) {
+				int l = is.read(b, 0, b.length);
+				if (l < 0)
+					break;
+				sb.append(new String(b, 0, l));
+			}
+			is.close();
```
alexeykazakov: is.close() should be in the finally {} block

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/util/BeanUtil.java
```diff
@@ -74,5 +75,20 @@ public static String getPropertyName(String methodName) {
...
+		int lastDotPosition = className.lastIndexOf("."); //$NON-NLS-1$
+		if(lastDotPosition >= 0 && className.length() > lastDotPosition){
+			className = className.substring(+1);
```
alexeykazakov: I guess you meant className = className.substring(lastDotPosition + 1);
Not just "+ 1".

And we should add a unit test for this method.

PR #398, opened 2015-04-28
JBIDE-17103 - allow feedback / return value / coreexception if runtime d...

robstryker:
...etection fails.  change required some small API changes

### DIFF
--- a/runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/model/AbstractRuntimeDetectorDelegate.java
+++ b/runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/model/AbstractRuntimeDetectorDelegate.java
@@ -11,10 +11,12 @@
 package org.jboss.tools.runtime.core.model;
 
 import java.io.File;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
+import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.jboss.tools.runtime.core.RuntimeCoreActivator;
 import org.jboss.tools.runtime.core.internal.RuntimeDetector;
@@ -22,10 +24,35 @@ import org.jboss.tools.runtime.core.internal.RuntimeDetector;
 public abstract class AbstractRuntimeDetectorDelegate implements
 		IRuntimeDetectorDelegate {
 
-	@Override
+	private boolean loggedWarning = false;
+	
+	/**
+	 * The framework will no longer call initializeRuntimes(List<RuntimeDefinition> runtimeDefinitions)
+	 * It will instead call initializeRuntime(RuntimeDefinition runtimeDefinition)
+	 */
+	@Override @Deprecated
 	public void initializeRuntimes(List<RuntimeDefinition> runtimeDefinitions) {
 	}
 
+	@Override
+	public boolean initializeRuntime(RuntimeDefinition runtimeDefinition) throws CoreException {
+		// Provide some migration code, for now.  This code may be removed at any time in the future. 
+		
+		// Since this is a new method, and some clients might not be implementing it, we 
+		// should redirect to the original method initializeRuntimes.
+		initializeRuntimes(Collections.singletonList(runtimeDefinition));
+		
+		// Log an error / warning indicating other handlers should update their API call
+		if( !loggedWarning ) {
+			RuntimeCoreActivator.pluginLog().logWarning("Runtime Detector " + findMyDetector().getId() + " is using a deprecated API call.");
+			loggedWarning = true;
+		}
+		
+		// We don't know what to return, since we don't know if anything was created
+		return false;
+	}
+
+	
 	@Override
 	public RuntimeDefinition getRuntimeDefinition(File root,
 			IProgressMonitor monitor) {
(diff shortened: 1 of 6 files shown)


robstryker:
There are two commits here, one in server and one in servertools. 

This change requires API changes. 

1) IRuntimeDetectorDelegate  has received a new method
2) This isn't a large breakage because AbstractRuntimeDetectorDelegate has implemented it, and clients are encouraged to extend AbstractRuntimeDetectorDelegate
3) the new method initializeRuntime(singleDefinition) includes logic to call the pre-existing method, but also log an error. If clients properly migrate, no error is logged. If there are any clients I missed, behavior should be the same as before, but an error will be logged indicating the code needs to migrate. 
4) RuntimeInitializerUtil.initializeRuntimes has had it's return type changed. This should not affect anyone directly, since it previously returned void. This does break binary compatibility, but is still source compatible, so clients will need to recompile against the new version. 


That's a summary of the changes.

runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/util/RuntimeInitializerUtil.java
```diff
@@ -39,13 +42,27 @@ public static void initializeRuntimesFromFolder(File directory, IProgressMonitor
...
+					}
+				}
+			}
+			
+			if( !found ) {
+				// Somehow display this error
+				ms.add(RuntimeCoreActivator.statusFactory().errorStatus("All runtime detectors failed to initialize " + def.getName()));
 			}
 		}
+		return ms;
```
fbricon: mmm a multistatus is returned even if no error occured. Should be OK status

runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/model/AbstractRuntimeDetectorDelegate.java
```diff
@@ -22,11 +24,32 @@
 public abstract class AbstractRuntimeDetectorDelegate implements
 		IRuntimeDetectorDelegate {
 
-	@Override
+	private boolean loggedWarning = false;
+	
+	@Override @Deprecated
```
fbricon: Needs to document which method to implement instead

runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/model/AbstractRuntimeDetectorDelegate.java
```diff
@@ -22,11 +24,32 @@
...
 
 	@Override
+	public boolean initializeRuntime(RuntimeDefinition runtimeDefinition) throws CoreException {
+		// Provide some migration code, for now.  This code may be removed at any time in the future. 
+		
+		// Since this is a new method, and some clients might not be implementing it, we 
+		// should redirect to the original method initializeRuntimes.
+		initializeRuntimes(Collections.singletonList(runtimeDefinition));
+		
+		// Log an error / warning indicating other handlers should update their API call
```
fbricon: this is wrong. the error will be logged when calling initializeRuntime(RuntimeDefinition runtimeDefinition) . you have a chicken/egg issue here :-)

runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/util/RuntimeInitializerUtil.java
robstryker: A multi-status with no children has a getSeverity() of OK.

runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/model/IRuntimeDetectorDelegate.java
```diff
@@ -29,9 +30,20 @@
 	 * the runtimes provided in this list. 
 	 * 
 	 * @param runtimeDefinitions
+	 * @deprecated
```
fbricon: document preferred method

runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/model/AbstractRuntimeDetectorDelegate.java
robstryker: > the error will be logged when calling initializeRuntime(RuntimeDefinition runtimeDefinition) 

This is intentiona.  The error will be logged when calling initializeRuntime(RuntimeDefinition runtimeDefinition)  only when a subclass of AbstractRuntimeDetectorDelegate has not yet implemented the new method.  If a subclass of AbstractRuntimeDetectorDelegate has implemented initializeRuntime(RuntimeDefinition runtimeDefinition), then it will not be using this method and no error will be logged.

fbricon:
If runtime detection version bumped to major, I'd remove the deprecated method directly

robstryker:
Removing deprecated methods entirely will break builds if others have not yet converted. Since the old method would be removed, there wouldn't even be an obvious place to document what new API the clients should be using. When making something API, I have a strong preference that when I deprecate it, I leave it there with proper errors / warnings until a full release has passed. 

The clients need fair warning and simply removing methods from interfaces is unacceptable. 

I think it's unfair to clients to remove methods without fair warning.

fbricon:
OK to apply once the deprecated methods have been documented

PR #401, opened 2015-05-12
JBIDE-19282 Unable to add HTML5 Palette tag when JS/CSS references ad…

dazarov:
…ded from JQuery Mobile palette

Added correction of offset and length before changing the document


common/plugins/org.jboss.tools.common.model.ui/src/org/jboss/tools/common/model/ui/views/palette/PaletteInsertHelper.java
```diff
@@ -138,6 +138,12 @@ protected void insertIntoEditorInternal(IDocument doc, Properties p, IPositionCo
 						getLineDelimiter(doc, firstLine - 1).length();
 			}
 			
+			if(start < 0){
+				start = 0;
+			}else if(start > doc.getLength()){
+				start = doc.getLength();
+			}
+
```
alexeykazakov: We cannot just use 0 or EOF as a position for the tag. It's not a solution if we insert the tag in some wrong place. We should find out why the position is not correct here.

PR #406, opened 2015-05-22
JBIDE-19500 warn/inform users running on non-compatilbe JVMs they are…

dazarov:
… loosing out on features

### DIFF
--- a/foundation/features/org.jboss.tools.foundation.feature/feature.xml
+++ b/foundation/features/org.jboss.tools.foundation.feature/feature.xml
@@ -31,7 +31,12 @@
          version="0.0.0"
          unpack="false"/>
 
-
+  <plugin
+         id="org.jboss.tools.foundation.checkup"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
 
 
 </feature>
--- a/foundation/features/org.jboss.tools.foundation.test.feature/feature.xml
+++ b/foundation/features/org.jboss.tools.foundation.test.feature/feature.xml
@@ -23,5 +23,11 @@
          install-size="0"
          version="0.0.0"
          unpack="false"/>
+   <plugin
+         id="org.jboss.tools.foundation.checkup.test"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
 
 </feature>
--- a/foundation/plugins/org.jboss.tools.foundation.checkup/.classpath
+++ b/foundation/plugins/org.jboss.tools.foundation.checkup/.classpath
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="output" path="bin"/>
+</classpath>
--- a/foundation/plugins/org.jboss.tools.foundation.checkup/.project
+++ b/foundation/plugins/org.jboss.tools.foundation.checkup/.project
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.jboss.tools.foundation.checkup</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.ManifestBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.jdt.core.javanature</nature>
+		<nature>org.eclipse.pde.PluginNature</nature>
+	</natures>
+</projectDescription>
--- a/foundation/plugins/org.jboss.tools.foundation.checkup/.settings/org.eclipse.jdt.core.prefs
+++ b/foundation/plugins/org.jboss.tools.foundation.checkup/.settings/org.eclipse.jdt.core.prefs
@@ -1,0 +1,12 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.methodParameters=do not generate
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.5
+org.eclipse.jdt.core.compiler.codegen.unusedLocal=preserve
+org.eclipse.jdt.core.compiler.compliance=1.5
+org.eclipse.jdt.core.compiler.debug.lineNumber=generate
+org.eclipse.jdt.core.compiler.debug.localVariable=generate
+org.eclipse.jdt.core.compiler.debug.sourceFile=generate
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
+org.eclipse.jdt.core.compiler.source=1.5
--- a/foundation/plugins/org.jboss.tools.foundation.checkup/META-INF/MANIFEST.MF
+++ b/foundation/plugins/org.jboss.tools.foundation.checkup/META-INF/MANIFEST.MF
@@ -1,0 +1,16 @@
+Manifest-Version: 1.0
+Bundle-Name: %Bundle-Name.0
+Bundle-ClassPath: .
+Bundle-Activator: org.jboss.tools.foundation.checkup.FoundationCheckupPlugin
+Bundle-Vendor: %providerName
+Bundle-SymbolicName: org.jboss.tools.foundation.checkup;singleton:=true
+Bundle-Localization: plugin
+Require-Bundle: org.eclipse.core.runtime;bundle-version="3.7.0",
+ org.eclipse.ui.workbench;bundle-version="3.107.0",
+ org.eclipse.ui.views.log;bundle-version="1.0.600",
+ org.eclipse.jface
+Bundle-Version: 1.2.0.qualifier
+Bundle-ActivationPolicy: lazy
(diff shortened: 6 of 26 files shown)


foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+import org.eclipse.ui.IStartup;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.internal.views.log.LogSession;
+import org.eclipse.ui.internal.views.log.TailInputStream;
+import org.eclipse.ui.progress.UIJob;
+
+
+public class JVMProblemDetector implements IStartup, ILogListener{
+	private static final String PREFERENCE_NAME="org.jboss.tools.foundation.core.JVMProblemDetector_NOT_TO_SHOW"; //$NON-NLS-1$
+	private static final String STRING_1 = "org.osgi.framework.BundleException: Could not resolve module:"; //$NON-NLS-1$
```
alexeykazakov: It's better not to hardoce these log messages but use the same strings/code which are used by osgi to generate these messages. So, if osgi change something in these messages then it will continue to be detected by our code. If it's possible of course.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+	
+	private UnresolvedStructure structure = new UnresolvedStructure();
+	
+	private LinkedBlockingDeque<String> queue = new LinkedBlockingDeque<String>();
+	
+	private String javaVersion = null;
+	
+	@Override
+	public void earlyStartup() {
+		PlatformUI.getPreferenceStore().setValue(PREFERENCE_NAME, false);
```
alexeykazakov: Do not forget to flush() the changed preferences. It will help to store preferences even if the eclipse session will be killed.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+		if(isAllowedToShow()){
```
alexeykazakov: This is always true, isn't it? You just set this preferences to "false" so isAllowedToShow() will always return "true" here.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+			while (true) {
+				String line0 = reader.readLine();
+				if (line0 == null)
+					break;
+				String line = line0.trim();
+				
+				scanLine(line);
+			}
+		} catch (FileNotFoundException e) { // do nothing
+		} catch (IOException e) { // do nothing
```
alexeykazakov: Why ignore IOException? Let's log it.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+		} finally {
+			try {
+				if (reader != null)
+					reader.close();
```
alexeykazakov: You can use a try-with-resources statement instead. Then you don't need to close the reader:
```java
try (BufferedReader reader=new BufferedReader(...)){
      ...
} catch (FileNotFoundException e) {
      ...
} catch (IOException e) { // do nothing
}
```

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
dazarov: Are you sure? There is no such thing as method flush() in IPreferenceStore, also I checked some references of IPreferenceStore.setValue() method - no one saves after this method.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
dazarov: You are right. I left this line by mistake

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
dazarov: Don't forget we are reading Error Log file. I am not sure if we cannot read file we could write in it

alexeykazakov: If you have any problem with reading the existing log you should log this.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
alexeykazakov: Ah, didn't notice you are using Workbench preference store instead of a IEclipsePreferences instance for your plugin.
You should not use the workbench preferences for storing your own internal plugin specific preferences. This storage has a lot of listeners and meant to use for publicly settable workbench preferences. And every time you change it you trigger them.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+				} finally {
+					try {
+						if (reader != null)
+							reader.close();
+					} catch (IOException e1) { // do nothing
+					}
+				}
+			}
+		} catch (InterruptedException e) {
+			e.printStackTrace();
```
alexeykazakov: Remove this e.printStackTrace();

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+					reader = new BufferedReader(new StringReader(message));
+					while (true) {
+						String line0 = reader.readLine();
+						if (line0 == null)
+							break;
+						String line = line0.trim();
+						scanLine(line);
+					}
+					structure.report();
+				} catch (IOException e) { // do nothing
```
alexeykazakov: The same as above. Empty catch and try-with-resources statement.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetectorMessages.properties
```diff
@@ -0,0 +1,17 @@
...
+### Distributed under license by Red Hat, Inc. All rights reserved.
+### This program is made available under the terms of the
+### Eclipse Public License v1.0 which accompanies this distribution,
+### and is available at http://www.eclipse.org/legal/epl-v10.html
+###
+### Contributors:
+###     Red Hat, Inc. - initial API and implementation
+##################################################################################
+
+SHOW_WARNING_DIALOG=Show warning dialog
```
alexeykazakov: Showing incompatible JRE version dialog

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+	}
+	
+	private void processStatus(IStatus status){
+		Throwable exception = status.getException();
+		String message = status.getMessage();
+		if(message != null){
+			try {
+				queue.put(message);
+			} catch (InterruptedException e) {
+				e.printStackTrace();
```
alexeykazakov: Remove e.printStackTrace();

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+	
+	public class UnresolvedStructure{
+		private List<UnresolvedModule> modules = new ArrayList<UnresolvedModule>();
+		
+		private List<String> moduleList = new ArrayList<String>();
+		private String moduleName = null;
+		
+		public UnresolvedModule[] getUnresolvedModules(){
+			ArrayList<UnresolvedModule> list = new ArrayList<UnresolvedModule>();
+			for(UnresolvedModule module : modules){
```
alexeykazakov: modules should be synchronized here.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+				if(module.isNeedReport()){
+					list.add(module);
+				}
+			}
+			return list.toArray(new UnresolvedModule[]{});
```
alexeykazakov: Do we really need to convert this list to an array? Why not to return the list itself here?

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
```diff
@@ -0,0 +1,475 @@
...
+		public void clear(){
+			synchronized(modules){
+				modules.clear();
+				moduleList.clear();
+				moduleName = null;
+			}
+		}
+		
+		public void addFirstModule(String name){
+			moduleList.clear();
```
alexeykazakov: Synchronize by modules.

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/JVMProblemDetector.java
dazarov: org.eclipse.jface.viewers.ITreeContentProvider requires arrays

PR #409, opened 2015-05-26
JBIDE-11436: License Feature

mickaelistria:


fbricon:
It should rather live under foundation, which is upstream from commons. Else foundation won't be able to consume the license feature

robstryker:
+1 to what fred says.  Must be in foundation.

mickaelistria:
Patch was updated to move feature to foundation.

mickaelistria:
@robstryker @fbricon: Can you please review it?

foundation/features/org.jboss.tools.foundation.license.feature/feature.properties
```diff
@@ -0,0 +1,49 @@
+###############################################################################
+# Copyright (c) 2010-2012 Red Hat, Inc. and others.
```
nickboldt: Shouldn't that say 2010-2015?

foundation/features/org.jboss.tools.foundation.license.feature/feature.properties
```diff
@@ -0,0 +1,49 @@
...
+# contains externalized strings for feature.xml
+# "%foo" in feature.xml corresponds to the key "foo" in this file
+# java.io.Properties file (ISO 8859-1 with "\" escapes)
+# This file should be translated.
+
+# "featureName" property - name of the feature
+featureName=JBoss Tools License feature
+
+# "providerName" property - name of the company that provides the feature
+providerName=Red Hat JBoss Middleware
```
nickboldt: Are we intentionally changing this from "JBoss by Red Hat" ?

foundation/features/org.jboss.tools.foundation.license.feature/feature.properties
```diff
@@ -0,0 +1,49 @@
...
+
+# "updateSiteName" property - label for the update site
+updateSiteName=JBoss Tools Update Site
+devUpdateSiteName=JBoss Tools Development Update Site
+
+# "description" property - description of the feature
+description=Common UI extensions used by JBoss Tools plugins.
+
+copyright=Copyright (c) 2007-2015 Exadel, Inc and Red Hat, Inc.\n\
```
nickboldt: I think this should be just "Red Hat, Inc., and others."

foundation/features/org.jboss.tools.foundation.license.feature/feature_ja.properties
```diff
@@ -0,0 +1 @@
+providerName=Red Hat JBoss Middleware
```
nickboldt: See comment above. Shouldn't this be "JBoss by Red Hat" ?

site/category.xml
```diff
@@ -80,7 +79,9 @@
 	</feature>
 
 <!-- foundation -->
-
+	<feature id="org.jboss.tools.foundation.license.feature" version="0.0.0">
+		<category name="AllFeatures"/>
```
nickboldt: Is there a source feature, too?

nickboldt:
My suggested changes, including cleaning up site/category.xml and switching to the NEW license (new RH legal address): https://github.com/jbosstools/jbosstools-base/pull/420

foundation/features/org.jboss.tools.foundation.license.feature/feature.properties
```diff
@@ -0,0 +1,49 @@
...
+
+# "providerName" property - name of the company that provides the feature
+providerName=Red Hat JBoss Middleware
+
+# "updateSiteName" property - label for the update site
+updateSiteName=JBoss Tools Update Site
+devUpdateSiteName=JBoss Tools Development Update Site
+
+# "description" property - description of the feature
+description=Common UI extensions used by JBoss Tools plugins.
```
nickboldt: Whoops, copy and paste error here.

foundation/features/org.jboss.tools.foundation.license.feature/feature.properties
mickaelistria: Right.

foundation/features/org.jboss.tools.foundation.license.feature/feature.properties
mickaelistria: I believe that "Red Hat JBoss Middleware" is actually the name we have to use here.

foundation/features/org.jboss.tools.foundation.license.feature/feature.properties
mickaelistria: Ok.

site/category.xml
mickaelistria: No, we don't need one.
license features just like as templates for other features, they are not installed so the source feature is useless.

mickaelistria:
@nickboldt  Thanks for the review. I see you've consumed this commit in some other pull request. Should I amend this one anyway (that would require you to rebase or cherry-pick the other changes you've made in #420 ) ?

nickboldt:
I don't think there's any point in updating yours to mine... unless you can show that I have the wrong providerName, and that we're switching to the new one as part of this change.

mickaelistria:
I'm too lazy and not interested enough to search or ask whether "Red Hat JBoss Middleware" is better "JBoss by Red Hat" or not. So let's just use your PR, to move things forward, and maybe one day we'll have directives on this topic, and it will be great because we'll only have to change one feature.

mickaelistria:
Replaced with #420

maxandersen:
we already have directives: https://github.com/jbosstools/jbosstools-devdoc/blob/master/source/legal.adoc

JBoss by Red Hat is afaik what we used for all other provider names for the opensource components, thus curious to know why Nick believes it now has to be Red Hat Middleware ?  not saying it is wrong just wondering if there are some specific reason.

mickaelistria:
It's not Nick who was thinking that, but me. I remember I've been told some time (maybe in the brainwashing session when I joined), that the "JBoss by Red Hat" logo was deprecated, and should be replaced by the "Red Hat JBoss Middleware" name. I thought it was the same for the text.

nickboldt:
Thanks, Mickael. 

@maxandersen, I see no clarification on which providerName to use in https://github.com/jbosstools/jbosstools-devdoc/blob/master/source/legal.adoc -- are you pointing us there so we will "ping JBoss Tools lead (manderse@redhat.com) directly" ? If so, consider yourself pinged. :D

maxandersen:
JBoss by Red Hat is the name used for our providers.
Red Hat JBoss Middleware is a legal entity which we at least so far only refer to in license text. I would just keep using what we already got here for now.

PR #410, opened 2015-05-26
JBIDE-19847 : use short sha1-based file names instead of long sanitized url names

fbricon:


robstryker:
A quick review shows this shouldn't affect existing cached files really, so this is a +1

PR #411, opened 2015-05-26
JBIDE-19820 Batch feature makes wildfly import way slower

scabanovich:


maxandersen:
seems like major change - could you please put some details in the git commit message ?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
```diff
@@ -0,0 +1,71 @@
...
+				if (entry.getName().endsWith(".class")) {
+					try {
+						InputStream stream = jar.getInputStream(entry);
+						try {
+							indexer.index(stream);
+						} finally {
+							stream.close();
+						}
+                    } catch (IOException e) {
+                    	CommonCorePlugin.getPluginLog().logError(e);
```
maxandersen: if there is an ioexception should this not return something like null or an empty index ?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                }
+            }
+            return indexer.complete();
```
maxandersen: is this safe to call if exception was thrown ?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
```diff
@@ -0,0 +1,71 @@
...
+        } finally {
+            safeClose(jar);
+        }
+    }
+
+	public static DotName toDotName(String qualifiedName) {
```
maxandersen: why does this have to be public ?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.jar.JarFile;
+
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.Index;
+import org.jboss.jandex.Indexer;
+import org.jboss.tools.common.core.CommonCorePlugin;
+
+public class JandexUtil {
+
+	public static Index createJarIndex(File jarFile, Indexer indexer) throws IOException {
```
maxandersen: do we really want to expose jandex API's outside this bundle ?

should this not be encapsulated and just provide the methods/queries we like to expose ?

maxandersen:
And how much difference does this make ? on the wildfly import case and in general ? is it worth it memory and CPU wise ?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
scabanovich: For tests, at least.

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
scabanovich: Method Indexer.complete() does not declare any exceptions. Our practice is not to catch java.lang.Exception. Should we catch it here?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
scabanovich: An empty index in this case is returned by indexer.complete().

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
scabanovich: Jandex API is public. Why is it wrong to open it? Wrapping it we will have to create our own API and maintain it.

scabanovich:
In the wildfly import case, it makes a great difference. Also, in case of changing maven dependencies. Figures in JBIDE-19820 imply 40% time increase for the wildfly workspace import when Batch tool is enabled (fix of JBIDE-19825 reduces it to 20%). Jandex indexes a jar almost as fast as it takes to read it. Hence, while we do not keep in memory the created indices and discard them after needed checks, it makes no load on memory. With JDT, when we check types in a jar, JDT keeps info for checked types in memory, and it can be several distinct objects for the same jar in context of different projects.

In general, when workspace and the number of jars in class path are not big, there is no need to optimise the jar check. However, if we take for another example JBossTools, which has more plugins and dependencies than wildfly, and suppose we would like to use Batch and CDI in our development, then we will be the first to demand the optimization. Though, I import JBossTools into a new workspace only 2 or 3 times a year. Some changes in dependencies take place more often, but that usually involves only some of libraries. All said, we have to vote if we need the optimisation or not. May I add my +1, Jandex is created by JBoss, it is already used by JBoss for more than a year.

common/tests/org.jboss.tools.common.core.test/src/org/jboss/tools/common/core/test/JandexTest.java
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	@Test
+	public void testJar() throws Exception {
+		IFile f = project.getFile("lib/weld-se-1.1.10.Final.jar");
+		assertTrue(f.exists());
+		Index index = JandexUtil.createJarIndex(f.getLocation().toFile(), new Indexer());
+		assertNotNull(index);
+		int classes = index.getAllKnownSubclasses(JandexUtil.toDotName("java.lang.Object")).size();
+		System.out.println("Classes=" + classes);
```
alexeykazakov: Let's remove this system.out.println().

scabanovich: Done.

alexeykazakov:
Rebased and pushed to master.

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/jandex/JandexUtil.java
maxandersen: by this you'll need to export the jandex api to *any* client of common - meaning you become the "owner" of that API.

scabanovich: Can we hide what we do not wish to expose in this way:
public static Object createJarIndex(File jarFile, Object indexer)
The method will check indexer and if it is org.jboss.jandex.Indexer, then the method will return an instance of org.jboss.jandex.Index, otherwise it will return null or better throw IllegalArgumentException?
Jandex API is just a few methods, I would not like to wrap it in 'our' own API that we will have to change any time we need more of Jandex.

maxandersen: That does not hide it - why would we want a public API method to return just Object?

scabanovich: Ok, I did to new PR:  https://github.com/jbosstools/jbosstools-base/pull/414, https://github.com/jbosstools/jbosstools-javaee/pull/346 that incapsulate Jandex inside common and open two query methods that are used by Batch and CDI.

PR #413, opened 2015-06-02

dgolovin:


common/plugins/org.jboss.tools.common/src/org/jboss/tools/common/log/BaseUIPlugin.java
```diff
@@ -11,14 +11,21 @@
 package org.jboss.tools.common.log;
 
 import java.util.HashMap;
+import java.util.Map;
```
robstryker: The changes to this file seem like they should be undone. They're just changes to imports

runtime/plugins/org.jboss.tools.runtime.ui/META-INF/MANIFEST.MF
```diff
@@ -10,7 +10,8 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.core.resources,
  org.eclipse.ui.navigator;bundle-version="3.5.100",
  org.jboss.tools.foundation.core;bundle-version="1.1.0",
- org.jboss.tools.foundation.ui;bundle-version="1.1.0"
```
robstryker: I don't believe a dependency on tools.common is appropriate here. I think you probably forgot to remove it.  We are trying to do more work in foundation and less in common. This patch, itself, puts all of the new code in foundation.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -10,19 +10,55 @@
...
+import org.eclipse.swt.widgets.MenuItem;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.browser.IWebBrowser;
 import org.eclipse.ui.browser.IWorkbenchBrowserSupport;
+import org.jboss.tools.foundation.core.ecf.URLTransportUtility;
 import org.jboss.tools.foundation.core.plugin.log.StatusFactory;
+import org.jboss.tools.foundation.ui.internal.FoundationUIPlugin;
+
+import sun.net.ProgressMonitor;
```
maxandersen: that seems like a bad import?

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -46,7 +82,7 @@ public void checkedCreateInternalBrowser(String url, String browserId, String pl
 		}
 	}
 	
-	public void checkedCreateExternalBrowser(String url, String pluginId, ILog log) {
+	public static void checkedCreateExternalBrowser(String url, String pluginId, ILog log) {
```
maxandersen: why are we changing this API to be static instead of instance based ?

robstryker:
https://github.com/jbosstools/jbosstools-base/pull/419   has been committed.  This one is rejected since it's incomplete for the download-runtime usecase

PR #415, opened 2015-06-04
JBIDE-19943 checkup plugin give NPE in tests of webservices

dazarov:


alexeykazakov:
Pushed to master and 4.3.0.Beta1x

PR #418, opened 2015-06-24
JBIDE-20139 Labels at property editors with 'Browse' button are missing

scabanovich:


alexeykazakov:
Rebased and pushed to master.

PR #419, opened 2015-06-25

robstryker:
Provide systemProperty to skip browsertion of internal browser

JBIDE-19335 - cleanup and API enhancement to allow url providers


foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -10,19 +10,57 @@
...
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.browser.IWebBrowser;
 import org.eclipse.ui.browser.IWorkbenchBrowserSupport;
+import org.jboss.tools.foundation.core.FoundationCorePlugin;
+import org.jboss.tools.foundation.core.IURLProvider;
+import org.jboss.tools.foundation.core.ecf.URLTransportUtility;
 import org.jboss.tools.foundation.core.plugin.log.StatusFactory;
+import org.jboss.tools.foundation.ui.internal.FoundationUIPlugin;
+
+import sun.net.ProgressMonitor;
```
maxandersen: what is this progressmonitor doing here ?

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -37,7 +75,7 @@
 	 * @param log
 	 *            the log provider to log against if an error occurred.
 	 */
-	public void checkedCreateInternalBrowser(String url, String browserId, String pluginId, ILog log) {
+	public static void checkedCreateInternalBrowser(String url, String browserId, String pluginId, ILog log) {
```
maxandersen: why the move to static ?

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -66,4 +104,280 @@ public void openUrl(String url, IWebBrowser browser, String pluginId, ILog log)
...
+	 * @see org.eclipse.swt.browser.Browser
+	 */
+	public static Browser createBrowser(int style, Composite parent, int preferredBrowser) {
+		/*
+			// We can provide webkit in this way:
+			   String defaultBrowser = System.getProperty(Browser.PROPERTY_DEFAULTTYPE);
+			   boolean hasDefaultBrowser = defaultBrowser != null;
+			   System.getProperties().setProperty(Browser.PROPERTY_DEFAULTTYPE, "webkit");
+		*/
+		if(System.getProperty("skipBrowserCreation")!=null) {
```
maxandersen: what is this for ?
 who/what will use that and why is it not prefixed with something ? 

and why the mixed case system property ? we use lowercase normally.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -66,4 +104,280 @@ public void openUrl(String url, IWebBrowser browser, String pluginId, ILog log)
...
+
+						}
+
+						@Override
+						public void setTaskName(String name) {
+							// TODO Auto-generated method stub
+						}
+
+						@Override
+						public void setCanceled(boolean value) {
```
maxandersen: why not support for cancel?

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -66,4 +104,280 @@ public void openUrl(String url, IWebBrowser browser, String pluginId, ILog log)
...
+						public void internalWorked(double work) {
+							// TODO Auto-generated method stub
+						}
+
+						@Override
+						public void done() {
+							link.getDisplay().asyncExec(new Runnable() {
+								@Override
+								public void run() {
+									// link
```
maxandersen: this does nothing?

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -66,4 +104,280 @@ public void openUrl(String url, IWebBrowser browser, String pluginId, ILog log)
...
+				return null;
+			}
+		};
+
+	}
+
+	private static Menu createLinkDefaultContextMenu (Control link, String url) {
+		Menu popupMenu = new Menu(link);
+	    MenuItem refreshItem = new MenuItem(popupMenu, SWT.NONE);
+	    refreshItem.setText("Messages.DownloadRuntimeLicenseFragment_Open_in_external_browser");
```
maxandersen: why is there anythign about runtime license fragment in the generic browser utility ?

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
robstryker: This was a change by Denis. I suspect he had a reason to do it.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
robstryker: This was a change by Denis. I suspect he had a reason to do it.   I will say, though, that the presence of the property has allowed me to test the behavior better and fix a bunch of bugs I would not have been able to replicate otherwise.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
robstryker: This method appears to be unused.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
robstryker: showProgressAndDownloadText  appears to be unused. I don't know what Denis' motivation was for putting it in.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
robstryker: Good catch.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
robstryker: Good catch, fixed by nick.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
robstryker: Verified:  before this patch, BrowserUtility had only one method in it, and was not static. It seems Denis' patch added a half-dozen or more methods, all of which were static, and this method was changed to static to be more uniform and for re-use by the newer signatures that were static and may have needed to use the existing method.

maxandersen: but..why? early on in foundation we agreed to avoid/reduce need for singletons in here.

Seems this brings no value, just limit its use IMO.

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
maxandersen: we should not let code into foundation without knowing why it is there.

if it is going to be there then follow the naming pattern/conventions we have elsewhere.

* lowercase
* "scoped" by package/feature

so that would be something like 'jbosstools.skip.browser.creation'  even though I can see it seems in foundation there is already several variations; but lets not make it worse.

and since it is a boolan it should not accept skipBrowserCreation=false

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
maxandersen: then please do not include it in foundation until we know.

PR #421, opened 2015-07-08
JBIDE-20184 Launching with Java 7 in jbdevstudio.ini results in module…

dazarov:
… errors in subsequent launches with Java 8

### DIFF
--- a/foundation/plugins/org.jboss.tools.foundation.checkup/src/org/jboss/tools/foundation/checkup/JVMProblemDetector.java
+++ b/foundation/plugins/org.jboss.tools.foundation.checkup/src/org/jboss/tools/foundation/checkup/JVMProblemDetector.java
@@ -14,10 +14,16 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.StringReader;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


foundation/plugins/org.jboss.tools.foundation.checkup/src/org/jboss/tools/foundation/checkup/JVMProblemDetector.java
```diff
@@ -39,6 +43,9 @@
 import org.osgi.framework.Constants;
 import org.osgi.service.prefs.BackingStoreException;
 
+import com.ibm.icu.text.DateFormat;
```
maxandersen: why is it using these from ibm.icu and not plain java versions ?

maxandersen:
looks "ok" but please add tests for checking it is sane.

alexeykazakov:
Rebased and pushed to master.

foundation/plugins/org.jboss.tools.foundation.checkup/src/org/jboss/tools/foundation/checkup/JVMProblemDetector.java
```diff
@@ -517,4 +547,99 @@ public int hashCode() {
...
+	/**
+	 * for test purpose
+	 * @param date string in yyyy-MM-dd HH:mm:ss.SSS 
+	 * @throws ParseException
+	 */
+	public void setEclipseStartTime(String str) throws ParseException{
+		eclipseStartTime = DATE_FORMATTER.parse(str).getTime();
+	}
+	
+	private long getEclipseStartTime() {
```
maxandersen: if you made this method protected you could just subclass it in tests to get the time you want and would not need to add a test only setEclipseStartTime method.

PR #425, opened 2015-07-20
JBIDE-19266 Batch validation problems QuickFixes

dazarov:
Code which detects the test environment moved from JVMProblemDetector to TestEnvironmentDetector

### DIFF
--- a/foundation/plugins/org.jboss.tools.foundation.checkup/src/org/jboss/tools/foundation/checkup/JVMProblemDetector.java
+++ b/foundation/plugins/org.jboss.tools.foundation.checkup/src/org/jboss/tools/foundation/checkup/JVMProblemDetector.java
@@ -81,20 +81,6 @@ public class JVMProblemDetector implements IStartup, ILogListener{
 	
 	private static final long MAX_LENGTH = 1024 * 1024;
 	
-	// possible values of eclipse.application system property
-	// need to cover:
-	// "org.eclipse.pde.junit.runtime.uitestapplication"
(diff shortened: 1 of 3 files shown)


alexeykazakov:
How is this related to Batch QuickFixes?

dazarov:
Batch QuickFixes Tests - the first tests which use this new approach to detect test environment

maxandersen:
eh...where do we have code that would want to have this kind of test ? 

that seems to be quite a codesmell ?

maxandersen:
i.e. before this patch the code was private and unused (i.e. dead code)
so i'm curious to know why we ever would want code that behaves differently based on what test runner it is in - that would be IMO.

dazarov:
We have tests for dialogs, which test how dialogs behave by default. So in tests dialog shoud not wait till user press buttons and finish button shoudl be presses automatically.
It's a matter of approach.
1. Have public method in dialogs which triggers test behaviour.
2. Have system property test sets it and dialog checks it.
3. Have general methos which detects that code was launchen in test environment...

#3 initially was made to test JVMProblemDetector, because #1 and #2 doesn't work where code is launched as early startup.

Now the main idea of using #3 is to have things done with less code and have universal method which can be used in different tests.
You don't need special test methods and you don't need to set system properties.

dazarov:
this PR depends on https://github.com/jbosstools/jbosstools-base/pull/425

alexeykazakov:
Rebased and pushed to master

PR #426, opened 2015-07-21
JBIDE-19335 Revert API Changes in BrowserUtility.java

dgolovin:


fbricon:
+1 to apply after checking we don't break more things.

PR #434, opened 2015-08-17
JBIDE-20447 2 Test Failure(s) in JBIDE 4.3.0.CR1 for common/jst/core …

kmarmaliykov:
…component

Making tests run in gtk2 mode


alexeykazakov:
It now fails locally on my ubuntu with GTK 3 :(

alexeykazakov:
I have merged the PR because it's more important to have working tests on Jenikins than on every local machine (only some linux machines are affected). But let's see what we can do on the jenkins side to fix that problem.

PR #441, opened 2015-08-21
JBIDE-20489 Performance issues of ELValidator and ELContextImpl

scabanovich:


fbricon:
Did you measure the performance gain?

scabanovich:
It depends on the number of ELs in a page. The gain is provided by replacing iteration by binary search. Thus, if there are less than 10 ELs, the gain is insignificant, for 100 ELs about it is 10 times faster. Also, improvement of this method goes together with improvements in jst (see JBIDE-20489) so that even for a small number ELs in a large page (more than 10Kb), page processing may be 10 times faster.

PR #444, opened 2015-09-03
JBIDE-20524: Adding useful API for OS detection

ibuziuk:


alexeykazakov:
Rebased and pushed to master.

maxandersen:
any reason this is not using things like eclipse Platform and/or EnvironmentInfo ? (http://help.eclipse.org/luna/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fcore%2Fruntime%2FPlatform.html)

ibuziuk:
@maxandersen well, according to the documentation it's not possible to detect windows 64 bit, cause 	*OS_WIN32* claims:
`Constant string (value "win32") indicating the platform is running on a Window 32-bit operating system (e.g., Windows 98, NT, 2000)`
However, it seems that it will return true even for 64 bit. Basically, inaccuracies in the docs put me off using that API

PR #448, opened 2015-09-05
JBIDE-20580 Performance: EL resolvers unnecessarily create a lot of empty ArrayList instances

scabanovich:


common/plugins/org.jboss.tools.common.el.core/src/org/jboss/tools/common/el/core/ca/AbstractELCompletionEngine.java
```diff
@@ -161,8 +166,11 @@ public IRelevanceCheck createRelevanceCheck(IJavaElement element) {
 	 * @see org.jboss.tools.common.el.core.resolver.ELResolver#resolve(org.jboss.tools.common.el.core.resolver.ELContext, org.jboss.tools.common.el.core.model.ELExpression)
 	 */
 	public ELResolution resolve(ELContext context, ELExpression operand, int offset) {
-		List<Var> vars = new ArrayList<Var>();
+		List<Var> vars = EMPTY_VARS;
 		Var[] array = context.getVars(offset);
+		if(array.length > 0) {
```
maxandersen: why not just:

vars = Arrays.asList(array); ?

scabanovich: Yes, it may be ok in this case. Arrays.asList should be used carefully, as it returns fixed size list - attempts to add/remove an element will result in exception.

common/plugins/org.jboss.tools.common.el.core/src/org/jboss/tools/common/el/core/ca/AbstractELCompletionEngine.java
```diff
@@ -161,8 +166,11 @@ public IRelevanceCheck createRelevanceCheck(IJavaElement element) {
 	 * @see org.jboss.tools.common.el.core.resolver.ELResolver#resolve(org.jboss.tools.common.el.core.resolver.ELContext, org.jboss.tools.common.el.core.model.ELExpression)
 	 */
 	public ELResolution resolve(ELContext context, ELExpression operand, int offset) {
-		List<Var> vars = new ArrayList<Var>();
+		List<Var> vars = EMPTY_VARS;
 		Var[] array = context.getVars(offset);
+		if(array.length > 0) {
+			vars = new ArrayList<Var>();
+		}
 		for (int i = 0; i < array.length; i++) {
```
fbricon: should be moved in previous if block

common/plugins/org.jboss.tools.common.el.core/src/org/jboss/tools/common/el/core/ca/AbstractELCompletionEngine.java
```diff
@@ -161,8 +166,11 @@ public IRelevanceCheck createRelevanceCheck(IJavaElement element) {
 	 * @see org.jboss.tools.common.el.core.resolver.ELResolver#resolve(org.jboss.tools.common.el.core.resolver.ELContext, org.jboss.tools.common.el.core.model.ELExpression)
 	 */
 	public ELResolution resolve(ELContext context, ELExpression operand, int offset) {
-		List<Var> vars = new ArrayList<Var>();
+		List<Var> vars = EMPTY_VARS;
 		Var[] array = context.getVars(offset);
+		if(array.length > 0) {
+			vars = new ArrayList<Var>();
```
fbricon: new ArrayList<Var>(array.length)

common/plugins/org.jboss.tools.common.el.core/src/org/jboss/tools/common/el/core/ca/AbstractELCompletionEngine.java
```diff
@@ -131,8 +133,11 @@ protected ImageDescriptor getELProposalImage(MemberPresentation memberPresentati
 	public List<TextProposal> getProposals(ELContext context, String el, int offset) {
 		List<TextProposal> completions = new ArrayList<TextProposal>();
 
-		List<Var> vars = new ArrayList<Var>();
+		List<Var> vars = EMPTY_VARS;
 		Var[] array = context.getVars(offset);
+		if(array.length > 0) {
+			vars = new ArrayList<Var>();
```
fbricon: new ArrayList(array.length)

common/plugins/org.jboss.tools.common.el.core/src/org/jboss/tools/common/el/core/ca/AbstractELCompletionEngine.java
```diff
@@ -188,10 +196,13 @@ public ELResolution resolve(ELContext context, ELExpression operand, int offset)
 	 * @return
 	 */
 	public ELResolution resolveELOperand(ELExpression operand, ELContext context, boolean returnEqualedVariablesOnly, int offset) {
-		List<Var> vars = new ArrayList<Var>();
+		List<Var> vars = EMPTY_VARS;
 		Var[] array = context.getVars(offset);
-		for (int i = 0; i < array.length; i++) {
-			vars.add(array[i]);
+		if(array.length > 0) {
+			vars = new ArrayList<Var>();
```
fbricon: new ArrayList(array.length)

scabanovich:
I consider adding method ELContext.getVarsAsList() that will return List<Var> and in this way there will be no making auxiliary lists that are transformed into arrays only to be put into new lists.

common/plugins/org.jboss.tools.common.el.core/src/org/jboss/tools/common/el/core/ca/AbstractELCompletionEngine.java
maxandersen: My suggested line does not use the asList directly, its wrapped in new ArrayList to avoid this issue.

Point was simply that your manual iteration of the array is exactly what asList does anyway.

And yeah, unnecessary objects can be an issue too but I don't know the code enough to tell.

Just spotted the many repeated for loops which seemed unnecessary and easily fixed.

scabanovich: I agree, this code needs further improvement.

PR #449, opened 2015-09-10
JBIDE-20580 Performance: EL resolvers unnecessarily create a lot of empty ArrayList instances

scabanovich:
ELContext.getVarsAsList(int) method is added to prevent creating
auxiliary lists transformed into arrays and then put into lists again.


alexeykazakov:
rebased and pushed to master

PR #451, opened 2015-09-17
JBIDE-20217 - rogue npe, adds a check but does not resolve root cause

robstryker:
This patch ensures an NPE doesn't blow away the stack, but it does not solve the root cause of why there's an NPE. It's possible that the root cause is related to JBIDE-20510,  but for now, this NPE check will ensure that isDebugModel(etc) returns correctly in the event of a not-found vm model (ie, a vm model that's not found is not debugable).


robstryker:
To be clear, I cannot replicate the NPE, but I can imagine how it can be happening.

alexeykazakov:
Maybe we should log if model == null if it's not what we are expecting? Just not to hide the problem at all.

robstryker:
We experimented with logging it earlier, but the problem is we'd end up with tons of logs for things we didn't expect.  For example, the JMX tool still runs jps every so often in the background, and sometimes the jps command completes and remains in the view for a few seconds before its removed.  We ended up getting logs for processes that recently completed. 

At this point, logging in this particular execution path is not the correct thing to do.

alexeykazakov:
OK. Let's apply it for CR1 but let's also create a new issue for 4.3.x to investigate it more and try to find a root problem and fix it.

PR #453, opened 2015-09-25
JBIDE-20750 Slow Memory Leak in JSF tools

scabanovich:


alexeykazakov:
Pushed to master and 4.3.x

PR #454, opened 2015-10-02
Improve operations that require SSE model

dgolovin:
StructuredModelWrapper update to allow return result from command.
Fix contains new Command2<T> class with T execute(IStructuredModel model) method
to allow return an object from the command. That way tricks with passing return through
final local array variable are not required anymore.


maxandersen:
which jira is this for ?

dgolovin:
There is no JIRA for that, I was just experimenting with Java Generics to improve code quality and avoid mistakes related to releasing IStructuredModel.

alexeykazakov:
Resolved conflict, linked to the https://issues.jboss.org/browse/JBIDE-20928 and pushed to master: https://github.com/jbosstools/jbosstools-base/commit/79a3a4e1972da4b09ac2b07efa2bdf1d01a2dcb0

PR #455, opened 2015-10-07
JBIDE-12359 Quick fixes don't work in WTP XML Editor

dazarov:


alexeykazakov:
Any tests?

maxandersen:
also this looks like an issue that should be fixed in mars too.

PR #459, opened 2015-10-25
JBIDE-20845 - Unable to install breakpoint even if line number set to be generated

snjeza:
https://issues.jboss.org/browse/JBIDE-20845
Unable to install breakpoint even if line number set to be generated

### DIFF
--- a/runtime/plugins/org.jboss.tools.runtime.ui/META-INF/MANIFEST.MF
+++ b/runtime/plugins/org.jboss.tools.runtime.ui/META-INF/MANIFEST.MF
@@ -10,7 +10,8 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.core.resources,
  org.eclipse.ui.navigator;bundle-version="3.5.100",
  org.jboss.tools.foundation.core;bundle-version="1.1.0",
- org.jboss.tools.foundation.ui;bundle-version="1.1.0"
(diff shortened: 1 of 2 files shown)


maxandersen:
This cannot be right. removing the warning won't actually fix this issue of the breakpoint not being hit.

And we should not be hardwiring user preferences in our setup like this IMO.

robstryker:
@Max:   Kinda off-topic, but, would it be reasonable to use this same 
strategy to disable launchbar?  ;)

maxandersen:
On 26 Oct 2015, at 12:30, Rob Stryker wrote:

> @Max:   Kinda off-topic, but, would it be reasonable to use this same
> strategy to disable launchbar?  ;)

if necessary, yeah.

> On 10/26/2015 08:21 AM, Max Rydahl Andersen wrote:
>>
>> This cannot be right. removing the warning won't actually fix this
>> issue of the breakpoint not being hit.
>>
>> And we should not be hardwiring user preferences in our setup like
>> this IMO.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/jbosstools/jbosstools-base/pull/459#issuecomment-151116275>.
>>
>
>
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/jbosstools/jbosstools-base/pull/459#issuecomment-151258208


/max
http://about.me/maxandersen

PR #460, opened 2015-10-30
JBIDE-21034 - enablement error on new runtime detectors

robstryker:


fbricon:
Why treat saving of detectors in 2 methods? It would be simpler if you saved each detector with its state, in one big preference blob

robstryker:
I suppose I could come up with a way to ensure that the two separate behaviors are treated differently and stored in one preference key, but I don't really see value in that. I could switch it to two lines of csv, with the first line being included and the second line being excluded... but I honestly think that is much less clear. 

I think the proper solution is two keys.

robstryker:
Ok, I see what you mean. 

Unfortunately, changing the property from a csv to a properties structure wouldn't be backwards compatible...    If you opened a neon workspace to test it, and I change the structure of the key, and then you open again in a previous version, the previous version won't know how to read the new structure. 

It's designed this way for backwards compatibility.

maxandersen:
please lets not continue to add complexity and fixes to this without having tests for this basic functionality.

robstryker:
I've updated the patch with a testcase. All tests pass. The test checks both API and internal implementation.

robstryker:
Also I'm not sure I'd characterize this patch as "added complexity", since it's an internal implementation detail and was designed to maintain backwards compatibility. A second preference key, while not ideal, is the proper solution considering the past implementation.

maxandersen:
Let me rephrase that "please lets not continue to add change and fixes to this without having tests for this basic but critical functionality". The level of complexity does not really matter - we need to be better at adding tests into verify if we have broken things or not.

runtime/tests/org.jboss.tools.runtime.test/src/org/jboss/tools/runtime/handlers/TestHandler2.java
```diff
@@ -0,0 +1,94 @@
...
+		return ME;
+	}
+	
+	public TestHandler2() {
+		ME = this;
+		System.out.println("Inside TestHandler2 constructor: " + this.toString());
+		try {
+			throw new Exception();
+		} catch(Exception e) {
+			e.printStackTrace();
```
alexeykazakov: Why do need to print the stack trace in the constructor? And btw Thread.dumpStack(); would look cleaner I guess ;-)

runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/internal/RuntimeCorePreferences.java
```diff
@@ -34,31 +36,54 @@ public static RuntimeCorePreferences getDefault() {
 		String enabledDetectors = getPreferences().get(ENABLED_DETECTORS,null);
 		return enabledDetectors == null ? null : enabledDetectors.split(","); //$NON-NLS-1$
 	}
+	public String[] getDisabledRuntimeDetectors() {
+		String enabledDetectors = getPreferences().get(DISABLED_DETECTORS,null);
```
alexeykazakov: Very minor thing. It doesn't change anything but for code cleanliness sake: "enabledDetectors" -> "disabledDetectors".

alexeykazakov:
Besides a couple of cosmetic comments looks good to me.

runtime/tests/org.jboss.tools.runtime.test/src/org/jboss/tools/runtime/handlers/TestHandler2.java
robstryker: I wasn't aware sysouts were prohibited in test code ;)   Honestly, I copied the class from TestHandler1 and sometimes use sysouts for debugging purposes.

PR #461, opened 2015-10-30
JBIDE-21034 - enablement error on new runtime detectors

robstryker:


alexeykazakov:
We should add tests to the "master" version too (as you added them to https://github.com/jbosstools/jbosstools-base/pull/460 ;) )

PR #463, opened 2015-11-09
[JBIDE-21067] Error Log view is polluted with NullPointerExceptions

dgolovin:
IResource.getLocation() can return null and returned value should be
checked for null before continue


alexeykazakov:
Pushed to master and 4.3.x

PR #464, opened 2015-11-13
JBIDE-21036 Infinite cycle in ParametedType.isAssignableTo()

scabanovich:


alexeykazakov:
Pushed to master and 4.3.x

PR #470, opened 2015-12-16
JBIDE-21323 EL Knowledge Base Problem with enums

scabanovich:


common/tests/org.jboss.tools.common.el.core.test/src/org/jboss/tools/common/el/core/test/resolver/TypeInfoCollectorTest.java
```diff
@@ -132,6 +132,19 @@ public void testProperties() throws CoreException {
...
+		IType bean = jp.findType(className);
+		TypeInfoCollector.TypeInfo typeInfo = new TypeInfoCollector.TypeInfo(bean, null, false);
+		TypeInfoCollector collector = typeInfo.getTypeCollector(false, false);
+		MemberInfo status = getMethod(collector, "getStatus");
+		assertNotNull(status);
+		collector = status.getTypeCollector(false, false);
+		MemberInfo name = getMethod(collector, "name");
+		assertNotNull(name);
+	}
+
```
andrei-ivanov: What about SignOffStatus#getLabel()?

scabanovich: It should work as property 'label'. Does it not?

andrei-ivanov: Ah, sorry, seems I've copy/pasted the wrong message in the issue. I've updated the issue now.
The point of the 2nd example was to show that even a standard property (label) isn't detected in some cases.

alexeykazakov:
Pushed to master and 4.3.x

PR #471, opened 2015-12-16
JBIDE-21338 - simple npe check when loading credentials model

robstryker:
targeted to beta2

### DIFF
--- a/foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/credentials/internal/CredentialDomain.java
+++ b/foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/credentials/internal/CredentialDomain.java
@@ -60,7 +60,7 @@ public class CredentialDomain implements ICredentialDomain {
 
 		credentials = new HashMap<String, String>();
 		String usersList = pref.get(PROPERTY_USER_LIST, (String)null);
-		if( !usersList.isEmpty()) {
+		if( usersList != null && !usersList.isEmpty()) {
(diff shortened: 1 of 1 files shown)


fbricon:
+1

robstryker:
pushed after rebase

foundation/plugins/org.jboss.tools.foundation.core/src/org/jboss/tools/foundation/core/credentials/internal/CredentialDomain.java
```diff
@@ -60,7 +60,7 @@ public CredentialDomain(Preferences pref) throws BackingStoreException {
 
 		credentials = new HashMap<String, String>();
 		String usersList = pref.get(PROPERTY_USER_LIST, (String)null);
```
dgolovin: I now it is too late, but why not this:
``` java
...
		credentials = new HashMap<String, String>();
  		String usersList = pref.get(PROPERTY_USER_LIST, "");
		if( !usersList.isEmpty()) {
  			String[] users = usersList.split("\n");
  			for( int i = 0; i < users.length; i++ ) {
  				credentials.put(users[i], NOT_LOADED_PASSWORD);
...
```
no null checks, no extra ternary operator.

robstryker: I suppose that would have worked as well, but I prefer to actually know whether the string was null (ie not set) or empty (ie previously persisted as an empty list) so that I can alter behavior based on those two conditions if I need to later.  Anyway, it's already been pushed.

PR #472, opened 2015-12-21
JBIDE-21374 EL var is not resolved correctly when value is set with bean method returning an array

scabanovich:


alexeykazakov:
Rebased and pushed to master and 4.3.x

PR #473, opened 2015-12-22
JBIDE-21358 - ensure processIsRunning only throws exceptions when critical

robstryker:
Ensure processIsRunning only throws exceptions when critical

### DIFF
--- a/common/plugins/org.jboss.tools.common.jdt.debug/src/org/jboss/tools/common/jdt/debug/tools/ToolsCore.java
+++ b/common/plugins/org.jboss.tools.common.jdt.debug/src/org/jboss/tools/common/jdt/debug/tools/ToolsCore.java
@@ -16,6 +16,7 @@ import java.util.Properties;
 import java.util.Set;
 
 import org.eclipse.jdt.launching.IVMInstall;
+import org.jboss.tools.common.jdt.debug.tools.internal.IToolsConstants;
 import org.jboss.tools.common.jdt.debug.tools.internal.Tools;
 
 /**
@@ -94,8 +95,32 @@ public class ToolsCore {
 		return Tools.getInstance().invokeActiveVms(hostname);
 	}
 	
+	/**
+	 * This method was poorly named and is is instead desired to return whether a process
+	 * is able to be monitored or not. 
+	 * 
+	 * @param hostname
+	 * @param vmPid
+	 * @return
+	 * @throws ToolsCoreException
+	 */
+	@Deprecated
 	public static boolean processIsRunning(String hostname, int vmPid) throws ToolsCoreException {
-		return Tools.getInstance().getMonitoredVm(hostname, vmPid) != null;
+		return processIsMonitorable(hostname, vmPid);
+	}
+	
+	public static boolean processIsMonitorable(String hostname, int vmPid) throws ToolsCoreException {
+		// If these throw exceptions, we want them to be rethrown and logged
+		Object host = Tools.getInstance().invokeGetMonitoredHost(hostname);
+		Object vmId = Tools.getInstance().invokeVmIdentifier(String.format(IToolsConstants.VM_IDENTIFIRER, vmPid));
+		
+		try {
+			// We want to catch and ignore these failures and just return that the process is able to be monitored
+			Object monitoredVm = Tools.getInstance().invokeGetMonitoredVm(host, vmId);
+			return monitoredVm != null;
+		} catch(ToolsCoreException tce) {
+			return false;
+		}
 	}
 
 	public static String getJvmArgs(String hostname, int vmPid) throws ToolsCoreException {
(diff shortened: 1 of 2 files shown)


common/plugins/org.jboss.tools.common.jdt.debug/src/org/jboss/tools/common/jdt/debug/tools/ToolsCore.java
```diff
@@ -95,7 +96,17 @@ public Object getMonitoredVM() {
...
-		return Tools.getInstance().getMonitoredVm(hostname, vmPid) != null;
+		// If these throw exceptions, we want them to be rethrown
+		Object host = Tools.getInstance().invokeGetMonitoredHost(hostname);
+		Object vmId = Tools.getInstance().invokeVmIdentifier(String.format(IToolsConstants.VM_IDENTIFIRER, vmPid));
+		
+		// We want to catch and ignore these failures and just return that the process is not running
+		try {
+			Object monitoredVm = Tools.getInstance().invokeGetMonitoredVm(host, vmId);
+			return monitoredVm != null;
+		} catch(ToolsCoreException tce) {
```
dgolovin: This code sure would fix problem with extra logging on console but side effects are unpredictable, in most cases we would not be able to do support, because piece of useful information would be missing. For example invokeGetMonitoredVm(host,vmId) call could trigger 6 types of exceptions:
``` java
		} catch (InvocationTargetException t) {
		} catch (IllegalAccessException e) {
		} catch (IllegalArgumentException e) {
		} catch (NoSuchMethodException e) {
		} catch (SecurityException e) {
		} catch (ClassNotFoundException e) {
```
Only one InvocationTargeException related to linked issue other are different scenarios. For example it would probably throw ClassNotFoundException for trying to monitor java < 1.5 (not sure if they appear in the list though). So instead of showing the warning about supporting only Java 1.5+ we just say: "OMG! Process is not running", which would be false statement.

robstryker: The problem is that these exceptions, if logged, could potentially be logged every x seconds. In almost all of these cases, it is best to simply ignore the process that suffers from *ANY* of these exceptions. 

Only one method calls processIsRunning, and that is RemoteDebugActivator.getVmModelsUsingTools().   This method is meant to create a model of things we *can* monitor. It is no benefit to us to list or pay attention to things we cannot monitor. 

If you're suggesting we should include *all* processes that are running into our model, then that would require substantial changes throughout the rest of the code to then further check whether they can be monitored. 

I can add a comment that indicates processIsRunning() would be more correctly named processIsMonitorable() to be more clear.

dgolovin: I didn't track the code that far, but it is always better to catch specific exceptions instead of generic ones. I see it is common approach in jdt.debug which IMO should be fixed as different issue.
For this pull request I would add new method processIsMonitorable() as you suggested which would explain why possible exceptions are ignored.

dgolovin: processIsMonitorable() should not declare any exceptions in throws and

common/plugins/org.jboss.tools.common.jdt.debug/src/org/jboss/tools/common/jdt/debug/tools/ToolsCore.java
```diff
@@ -94,8 +95,32 @@ public Object getMonitoredVM() {
...
+	 */
+	@Deprecated
 	public static boolean processIsRunning(String hostname, int vmPid) throws ToolsCoreException {
-		return Tools.getInstance().getMonitoredVm(hostname, vmPid) != null;
+		return processIsMonitorable(hostname, vmPid);
+	}
+	
+	public static boolean processIsMonitorable(String hostname, int vmPid) throws ToolsCoreException {
+		// If these throw exceptions, we want them to be rethrown and logged
+		Object host = Tools.getInstance().invokeGetMonitoredHost(hostname);
```
dgolovin: Shouldn't we get these two lines  inside try/catch and remove ToolsCoreException from throws?

robstryker: No... I consider those two critical errors. If the running VM cannot locate the monitored host (ie localhost) or is incapable of properly using the sun.jvmstat.monitor.VmIdentifier constructor (which is very simple), then I'd see that as a critical error that should be re-thrown and logged. 

If however we can access the monitored host, and can use the sun.jvmstat.monitor.VmIdentifier constructor without error, but an exception is thrown on invokeGetMonitoredVM (where most of the work is done, ie, where we ask the host to locate and connect to the process), then that should be caught and ignored, because that is an expected failure. 

Basically, the expected failures are inside the try/catch and ignored. The unexpected failures are outside of the try/catch, thrown up the stack, and logged.

dgolovin:
I don really see the difference between those two, my understanding is we need to fill UI with Java processes we can monitor. In your differentiation some processes going to report exception some are not, but both still cannot be monitored anyway.

robstryker:
Being unable to locate the host (ie localhost) or the proper java class (class not found) are what I consider critical errors, because it means that *no* process can be found. This is something the user should be made aware of. In this case, it's not a specific process that cant be found... its that none can becuase these errors would occur for all processes being looked for. 

Being unable to locate the specific process could just mean it terminated 1ms ago and is thus an expected error.

fbricon:
Code looks good to me. Rob's argument makes sense to me

PR #474, opened 2015-12-23
JBIDE-21383 Content assist for facelet function does not work in an unclosed EL before next EL

scabanovich:


alexeykazakov:
Rebased and pushed to master and 4.3.x branch

PR #477, opened 2016-01-20
JBIDE-19178 Explorer, Servers view: Add deployment view to OpenShift v3 explorer

scabanovich:
JBoss perspective is modified: Properties view is put to the right of Problems+Servers+Openshift.


fbricon:
+1 to apply

fbricon:
applied in master/4.3.x

PR #478, opened 2016-01-21
Small api enhancement required for openshift in JBIDE-21481

robstryker:


runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/extract/ExtractUtility.java
```diff
@@ -41,6 +46,20 @@ public ExtractUtility(File file) {
...
+			case FORMAT_ZIP:
+				util = new UnzipUtility(file);
+				break;
+			case FORMAT_TAR:
+				util = new UntarUtility(file);
+				break;
+			case FORMAT_TGZ:
+				util = new UntarGZUtility(file);
+				break;
+		}
```
fbricon: what if it's not supported? is there a default strategy? should it throw an exception?

alexeykazakov:
Base has been already branched for Beta2. Don't forget to push your changed to 4.3.1.Beta2x too if you want them in Beta2.

runtime/plugins/org.jboss.tools.runtime.core/src/org/jboss/tools/runtime/core/extract/ExtractUtility.java
robstryker: The .extract() method will return an IStatus.  If it can't discover how to extract, it will return an error status. If it attempts to extract using the wrong format (ie unzip a tgz file) the underlying extraction will throw a CoreException, which ExtractUtility catches and wraps with an error status. 

So I think the rest of existing api there is sufficient for now.

fbricon:
+1 to apply in all branches

robstryker:
merged to all 3

PR #480, opened 2016-01-27
[JBIDE-21587] moved databinding utilities openshift -> common

adietish:


common/plugins/org.jboss.tools.common.ui/src/org/jboss/tools/common/ui/databinding/EmptyStringToNullConverter.java
```diff
@@ -0,0 +1,37 @@
...
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.common.ui.databinding;
+
+import org.eclipse.core.databinding.conversion.Converter;
+import org.jboss.tools.openshift.common.core.utils.StringUtils;
```
alexeykazakov: openshift? common doesn't depends on OS.

common/plugins/org.jboss.tools.common.ui/src/org/jboss/tools/common/ui/databinding/HostNameValidator.java
```diff
@@ -0,0 +1,40 @@
...
+import org.eclipse.core.databinding.validation.ValidationStatus;
+import org.eclipse.core.runtime.IStatus;
+import org.jboss.tools.openshift.common.core.utils.StringUtils;
+
+/**
+ * @author Andre Dietisheim
+ */
+public class HostNameValidator implements IValidator {
+
+	private static final Pattern urlPattern =
```
robstryker: This seems to duplicate SimpleUrlStringChecker.simpleUrlPattern

common/plugins/org.jboss.tools.common.ui/src/org/jboss/tools/common/ui/databinding/EmptyStringToNullConverter.java
adietish: good catch, correcting.

common/plugins/org.jboss.tools.common.ui/src/org/jboss/tools/common/ui/databinding/HostNameValidator.java
adietish: good catch, correcting.

adietish: Turns out the urlPattern is actually a host pattern, not an url pattern, thus not duplicate. Renaming the constant.

common/plugins/org.jboss.tools.common.ui/src/org/jboss/tools/common/ui/databinding/EmptyStringToNullConverter.java
adietish: replaced by org.apache.commons.lang.StringUtils

robstryker:
I'm a bit curious why we're putting stuff in 'common' still?  As I understood it, common is a set of plugins that are slightly legacy and suffer from a lot of problems, specifically mixing core and ui in it's dependency lists... and a severe lack of unit tests. 

If this set of classes is really useful enough to be re-used by JBT as a whole, then it should probably go in foundation.ui and have unit tests. 

There is no good reason at all to be adding code to o.j.t.common, in my opinion... but I'd welcome a discussion on the topic.

robstryker:
Also common has deps on stuff like WTP, which makes it not useful for plugins that don't want to force the environment to contain wtp or drag in random wtp plugins to potentially cause an incomplete environment.

alexeykazakov:
I have the same concerns as @robstryker. Maybe foundation is really better place for that stuff? @maxandersen wdyt?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/databinding/DirectoryValidator.java
```diff
@@ -0,0 +1,40 @@
...
+ ******************************************************************************/
+package org.jboss.tools.common.databinding;
+
+import java.io.File;
+
+import org.eclipse.core.databinding.validation.ValidationStatus;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.osgi.util.NLS;
+
+/**
```
maxandersen: in general we need to document what our classes do, but especially in commons ( or foundation)  we really need to have some API docs and accompanying tests.

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/databinding/IsNotNullValidator.java
```diff
@@ -0,0 +1,35 @@
...
+import org.eclipse.core.runtime.IStatus;
+
+/**
+ * @author Andre Dietisheim
+ */
+public class IsNotNullValidator implements IValidator {
+
+	private IStatus invalidStatus;
+
+	public IsNotNullValidator(IStatus invalidStatus) {
```
maxandersen: no docs for this istatus and also seem all other IValidators follows a different pattern.

What is the reason for the difference here ?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/databinding/MultiConverter.java
```diff
@@ -0,0 +1,90 @@
...
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.common.databinding;
+
+import org.eclipse.core.databinding.conversion.IConverter;
+import org.eclipse.osgi.util.NLS;
+
+/**
+ * A converter that invokes several given converters.
```
maxandersen: what is the purpose of that ?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/databinding/ObservablePojoUtils.java
```diff
@@ -0,0 +1,38 @@
...
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.common.databinding;
+
+import java.beans.PropertyChangeListener;
+
+/**
+ * @author Andre Dietisheim
+ */
+public class ObservablePojoUtils {
```
maxandersen: this looks like an anti pattern - why would one ever have a list of objects that one runs through where only some objects are observable ?

wouldn't that always be an error situation ?

common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/databinding/SimpleUrlStringChecker.java
```diff
@@ -17,7 +17,7 @@
  */
```
maxandersen: no docs?

common/plugins/org.jboss.tools.common.ui/src/org/jboss/tools/common/ui/databinding/FormPresenterSupport.java
```diff
@@ -0,0 +1,376 @@
...
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Listener;
+
+/**
+ * Connects the validation result from the given data binding context to the
+ * given wizard fragment, updating the page's error message accordingly.
+ *
+ * @since 1.3
```
maxandersen: is that the right version ?

maxandersen:
I've put comments on issues I could find on first run but don't think I got them all.

But there are not sufficient docs, and no tests (at all) for this. Thus it is not something we should put in Foundation before that is fixed and it really aren't good we keep piling that up into Commons either.

adietish:
this PR is 4y old and active development moved on to odo based tooling. Closing.

PR #482, opened 2016-01-28
JBIDE-21529 Properties view: Columns of OpenShift resources in tabbed properties are too wide

scabanovich:


foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/AutoResizeTableLayout.java
```diff
@@ -0,0 +1,129 @@
+/*************************************************************************************
+ * Copyright (c) 2010-2011 Red Hat, Inc. and others.
```
fbricon: 2010-1016

scabanovich:
This pull request should be rejected, it is not needed in new version of fix for JBIDE-21529.

PR #484, opened 2016-02-09
JBIDE-21302 Removing user from credentials framework while not authenticated to secure storage should not work

scabanovich:
Call CredentialService.getCredentialModel().saveModel() before removal is done.
That triggers authentication to secure storage if it has not been done yet.
By returning boolean in saveModel() equal to false if user failed to authenticate,
we can abort removing credentials.

### DIFF
--- a/foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/credentials/internal/CredentialPreferencePage.java
+++ b/foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/credentials/internal/CredentialPreferencePage.java
@@ -15,6 +15,7 @@ import java.util.Collections;
 import java.util.Comparator;
 
 import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.jface.dialogs.MessageDialog;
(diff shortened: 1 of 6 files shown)


foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/credentials/internal/CredentialPreferencePage.java
```diff
@@ -365,11 +373,20 @@ private void addUserPressed() {
 		}
 
 	}
+
+	private static final String UNABLE_TO_DELETE_CREDENTIALS = "Unable to delete credentials stored in secure storage without authenticating to secure storage";
```
robstryker: Please put this in a proper externalized class...   Specifically  CredentialMessages.java and CredentialMessages.properties. 

Otherwise, great patch.

scabanovich: Ok, done.

robstryker:
merged after a rebase. good job!

robstryker:
Outdated PR.

PR #486, opened 2016-02-11
JBIDE-21673 JBoss Runtime Detection preference page throws StackOverflowError

scabanovich:


fbricon:
+1 to apply in master and maintenance

PR #487, opened 2016-02-12
JBIDE-21674 Enable hierarchical view by default in Project Explorer

scabanovich:


mickaelistria:
A quick overview makes that this approach seems good: doesn't require any upstream change, and allows user to easily use the legacy flat project layout of they want to. So +1 on the design.
I won't have the opportunity to actually try that and get into details of the code before the 22nd. So if someone else validates it and there is a consensus that it is a good thing, +1 to merge it.

maxandersen:
without tests I'm -1 on this kind of change this late in the game.

Requiring our base/foundation plugins participating in startup and toggling these kind of settings is really not something that feels safe imo.

robstryker:
Very old issue and I agree with Max. Closing out.

PR #488, opened 2016-02-17
JBIDE-21707 NullPointerException in JobResultFuture

scabanovich:


common/plugins/org.jboss.tools.common.ui/src/org/jboss/tools/common/ui/JobResultFuture.java
```diff
@@ -41,12 +41,13 @@ public JobResultFuture(Job job) {
 
 	@Override
 	public boolean cancel(boolean mayInterruptIfRunning) {
-		if ((isRunning(job)
-				&& mayInterruptIfRunning)
-				|| !isRunning(job)) {
```
adietish: why cancel the job if it's not running?

scabanovich: It may be still waiting or sleeping. JobManager does all the checks and changes state to NONE. So, even if the current state of job is NONE, it does not harm to call cancel().
For the boolean expression, I only made it a bit shorter, but it is equivalent to what it was.

scabanovich: State 'sleeping' may mean UI jobs deadlock, the ability to cancel a sleeping job is the true way to salvation.

fbricon:
+1 LGTM

PR #490, opened 2016-03-01
JBIDE-21803 Dialog Add a Credential Domain allows create domain with empty name

scabanovich:


robstryker:
very clean and unexpected patch. Thanks a lot!   +1

Do we have any idea where the null name came from?  Is it really just that the wizard allows no name? Or do we suspect something wrong when persisting a valid object?

scabanovich:
New domain dialog initializes domainName with null. Then if user makes no input that null could be saved into model as dialog did not disabled finish. So, this usecase reproduced the problem with certainty. I could not reproduce the problem when creating valid domane names (which though does not prove that something cannot ever go wrong).

robstryker:
ok thanks for looking into this. You've been a big help.

PR #492, opened 2016-03-02
JBIDE-21795 - npe on save ina broken workspace (but still good to fix)

robstryker:


robstryker:
@scabanovich This might be a cause

scabanovich:
+1. Yes it is better to get visible name from id if the former is empty.

PR #496, opened 2016-03-11
JBIDE-21719 Move Properties view to its previous position

scabanovich:


common/plugins/org.jboss.tools.common.ui/src/org/jboss/tools/common/ui/JBossPerspectiveFactory.java
```diff
@@ -68,35 +73,39 @@ public void createInitialLayout(IPageLayout layout) {
...
+		bottomRight.addPlaceholder(add(IPageLayout.ID_BOOKMARKS, knownViews));
+		bottomRight.addPlaceholder(add(IProgressConstants.PROGRESS_VIEW_ID, knownViews));
+		bottomRight.addPlaceholder(add(ID_SEARCH_VIEW, knownViews));
 
 		// Top right.
 		IFolderLayout topRight = layout.createFolder("topRight", IPageLayout.RIGHT, 0.7f, editorArea);//$NON-NLS-1$
-		topRight.addView(IPageLayout.ID_OUTLINE);
+		topRight.addView(add(IPageLayout.ID_OUTLINE, knownViews));
+
+		assignAllUnknownViewsToDefaultFolder(bottomRight, knownViews);
```
alexeykazakov: So, it will assign all unknown views to the bottom right? Isn't it the default eclipse behavior which you want to override?

scabanovich: We define also 'bottomRight2' which goes to the right of 'bottomRight'. By default Eclipse drops unknown views to the most right at bottom, which is  'bottomRight2' in this case, and we do not want that. In JBoss perspective 'bottomRight' is in the center.

robstryker:
Patch out of date

PR #498, opened 2016-03-17
JBIDE-21894 NPE when canceling openshift new app wizard

scabanovich:


scabanovich:
@adietish , @fbricon , please try the pull request. Since most wizards open through WizardUtils.openWizardDialog(), a couple of other wizards apart from new app wizard need be checked too, to make sure they open in spite of the cancelling condition.

PR #502, opened 2016-04-29
[JBIDE-20937] Create resource(s) from json

jeffmaury:
- Add UrlValidator and sub-classes from Apache Common Validator

### DIFF
--- a/common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/validation/InetAddressValidator.java
+++ b/common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/validation/InetAddressValidator.java
@@ -1,0 +1,184 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.tools.common.core.validation;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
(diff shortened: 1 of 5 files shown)


common/plugins/org.jboss.tools.common.core/src/org/jboss/tools/common/core/validation/InetAddressValidator.java
```diff
@@ -0,0 +1,184 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
alexeykazakov: Why Apache licence?
All our code should be under EPL with Red Hat copyright:

/******************************************************************************* 
 * Copyright (c) 2016 Red Hat, Inc. 
 * Distributed under license by Red Hat, Inc. All rights reserved. 
 * This program is made available under the terms of the 
 * Eclipse Public License v1.0 which accompanies this distribution, 
 * and is available at http://www.eclipse.org/legal/epl-v10.html 
 * 
 * Contributors: 
 * Red Hat, Inc. - initial API and implementation 
 ******************************************************************************/

scabanovich: Is it http://commons.apache.org/proper/commons-validator/ ? If so, can we just include it into classpath like
<dependency>
	<groupId>commons-validator</groupId>
	<artifactId>commons-validator</artifactId>
	<version>1.4.0</version>
</dependency>

jeffmaury: Yes, it is code copied from Apache Validator and Fred told me this was allowed on a case by case because I can't find this artifact in the target platform

scabanovich: Plugin can have in pom.xml a task that downloads the jar into folder 'lib' (registered in .gitignore) and .classpath will reference the jar. As an example, plugin org.jboss.tools.openshift.client downloads some jars in this way.

jeffmaury: I know but that's an OSGI anti-pattern

scabanovich: Target platform can be modified later, then just few lines have to be removed from pom.xml, .gitignore and .classpath. 
With copied classes, we will have to remove all of them and do refactoring of imports.

alexeykazakov: So you didn't modify the original code (besides changing package name)? I'm not sure this is the best solution. If we copy sources just to use them as is then we will have to maintain this code etc.

jeffmaury: No but I will close this PR and use the dependencies and lib approach

PR #506, opened 2016-05-19
JBIDE-22379 Server editor: closing it before the widgets are enabled ends up in "widget disposed" error

scabanovich:


scabanovich:
The issue was fixed by a change in openshift. However, this pull request remains actual, as class DisableAllWidgetsJob is API and can be reused by another component.

robstryker:
pushed after rebase

PR #510, opened 2016-06-01
JBIDE-22458 : JBoss Help should use DCP v2

fbricon:


foundation/plugins/org.jboss.tools.foundation.help.ui/src/org/jboss/tools/foundation/help/ui/internal/search/SearchiskoResultBuilder.java
```diff
@@ -77,11 +80,20 @@ public String toAbsoluteHref(String href, boolean frames) {
 		}
```
robstryker: Please document behavior, that if the type is a list, it uses first item in the list.

robstryker:
Other than documenting the new behavior, seems +1-able.

PR #513, opened 2016-06-28
Fix test errors on systems without browser available

dgolovin:
If browser is not available test should check for Link instance
as well in latest check


alexeykazakov:
This PR fixes testCreateBrowserOrLink() but testCreateBrowser() test will fail anyway.

dgolovin:
That is true. On the other side there is no reason to have testCreateBrowser(). testCreateBrowserOrLink() triggers createBrowser() call as well.

PR #514, opened 2016-06-29
JBIDE-13671 set correct BUILD_ALIAS in...

nickboldt:
JBIDE-13671 set correct BUILD_ALIAS in plugin version of foundation.core so when we switch to jgit timestamps this plugin will still always be newer

### DIFF
--- a/foundation/plugins/org.jboss.tools.foundation.core/pom.xml
+++ b/foundation/plugins/org.jboss.tools.foundation.core/pom.xml
@@ -23,24 +23,33 @@
 			</resource>
 		</resources>
 		<plugins>
-      <!-- get major.minor.incremental from root pom, then use devstudio.version = ${parsedVersion.majorVersion}.${parsedVersion.minorVersion}.${parsedVersion.incrementalVersion}.${BUILD_ALIAS} -->
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>build-helper-maven-plugin</artifactId>
-        <version>1.9.1</version>
-        <executions>
-          <execution>
-            <phase>validate</phase>
-            <id>parse-version</id>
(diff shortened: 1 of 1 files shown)


robstryker:
For documentation purposes:

Nick:  all I'm doing is moving a piece of logic from parent pom down to foundation.core and devstudio.core.central because only those TWO plugins will continue to use a BUILD_ALIAS in their plugin names
---

Seems +1able to me.

nickboldt:
Note that this change is really just copying the same logic in parent pom to this pom. See

https://github.com/jbosstools/jbosstools-build/blob/d0db3e076b2b55cf2bce3b5713c681ea31704076/parent/pom.xml#L151
and
https://github.com/jbosstools/jbosstools-build/blob/d0db3e076b2b55cf2bce3b5713c681ea31704076/parent/pom.xml#L691

Once this is applied, we can then apply https://github.com/jbosstools/jbosstools-build/pull/223 to remove the logic in parent pom and move to jgit timestamps

nickboldt:
Awesome. PR build failed one single test, so reports "all the checks have failed".

https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/jbosstools-base-Pull-Request/15/testReport/junit/org.jboss.tools.runtime.test/RuntimeDetectionFrameworkTest/testLoadSaveRuntimePaths/

nickboldt:
retest this please

nickboldt:
retest this please

alexeykazakov:
Looks good to me.
@dgolovin could you please take a look too?

dgolovin:
Build locally with all related PR's and it seems to work as expected.

nickboldt:
Cool. Can I get a PR-merge witness?

PR #515, opened 2016-06-29
JBIDE-22651 - fix favicon caching on windows

robstryker:


robstryker:
I'm really not sure this is the best patch possible, but it's what I have to work with. 
For some reason, java utility code  Files.probeContentType(f.toPath());    fails against a .tmp file. 

I've looked online for a long while to try to see what's causing it, but I can't. Some of it might be registry-based, and there are multiple reports of the default mechanism of this API failing on Mac and Windows in various configurations.  So it's not really clear if anyone is keeping this API up to date over at oracle. 

This patch actually changes the downloading mechanism code to not use .tmp suffix in all downloads, but rather to use a suffix if the url in question has an obvious suffix of length 2 to 4.  So if the url ends in .JPEG,  the downloaded file will also end in .JPEG. 

Somehow, Files.probeContentType(f.toPath()); returns the correct mime-type if the suffix is correct. 

Unfortunately, this patch will still not behave properly if the remote get returns a 404 or redirect to some html page. In fact, this patch is incomplete because the UI will basically get stuck in an endless loop. Damnit.

robstryker:
Ok. no... this patch is bad.  The windows implementation will 100% trust the suffix. So if it's a 404 / html page stored in a .ico file, the mime-type check will still claim it's an image/ico file, despite it being a plain text file. 

Crap.

robstryker:
I've looked into ImageIO  class in java, but it doesn't support or recognize .ico files, so will not work for favicon. 

ImageIO.read(file) - doesn't work, will not read .ico files
MimetypeFileTypesMap.getContentType(file) - seems to respond based on suffix
Files.probeContentType(file) - also seems to work based on suffix?

alexeykazakov:
Seems to be outdated. Closing.

PR #519, opened 2016-07-22
JBIDE-22651 - Importing magicfile4j library for detecting file types to

robstryker:
fix credential UI bug in downloading favicons


robstryker:
Right now, the library is included directly as an internal implementation detail of foundation.ui. 

I think it might be wise to eventually break it out into it's own plugin so that others can access the API or use it. But for this release, I think it's acceptable to keep it as an internal detail.

robstryker:
retest this please

jeffmaury:
Why do you need a lib where FileTypeMap can do the job ?

PR #531, opened 2016-09-13
fix currentversion.properties and plugin...

nickboldt:
fix currentversion.properties and plugin version of org.jboss.tools.foundation.core


nickboldt:
Replaced by #532

PR #535, opened 2016-10-12
JBIDE-23302 if we don't actually need it, we...

nickboldt:
JBIDE-23302 if we don't actually need it, we could switch to http instead of http for the AERI discovery/redirect to https://aer.ctrlflow.com/redhat/community/discovery


maxandersen:
it is always recommended to use https these days to avoid man-in-the-middle attacks or snoopy proxy/firewalls.

robstryker:
Hi Nick:

Any advice on what this is for or what needs to be tested to ensure this works?

robstryker:
Outdated and bad patch.

PR #538, opened 2016-10-17
JBIDE-23381 - split the monolith

robstryker:


robstryker:
https://github.com/jbosstools/jbosstools-base/pull/538/files#diff-97c3f32a706eb744a5fd1051e2837c0fL33   is the real work of moving the dep to foundation.ui from core. 

The rest is splitting out inner classes and trying to make sure references to the model are passed around, rather than having a monolith of inner classes that's impossible to understand.

PR #542, opened 2016-10-27
JBDS-4014 Add a facility to pick up and report on RPM (or any OS package

mbooth101:
manager) installed packages. Product ID is sent back augmented
with a comma delimited list of such packages.

To make use of this, the OS installed packages that we are
interested in must deliver a file into $eclipse_home/.pkgs
as part of its payload.

### DIFF
--- a/usage/tests/org.jboss.tools.usage.test/META-INF/MANIFEST.MF
+++ b/usage/tests/org.jboss.tools.usage.test/META-INF/MANIFEST.MF
@@ -10,7 +10,8 @@ Require-Bundle: org.jboss.tools.usage;bundle-version="1.0.0",
  org.eclipse.osgi;bundle-version="3.5.2",
  org.eclipse.core.runtime;bundle-version="3.5.0",
  org.junit;bundle-version="[4.8.1,5.0.0)",
- org.eclipse.jface;bundle-version="3.10.0"
+ org.eclipse.jface;bundle-version="3.10.0",
+ org.hamcrest.core;bundle-version="1.3.0"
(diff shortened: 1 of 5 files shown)


usage/tests/org.jboss.tools.usage.test/META-INF/MANIFEST.MF
```diff
@@ -10,7 +10,8 @@ Require-Bundle: org.jboss.tools.usage;bundle-version="1.0.0",
  org.eclipse.osgi;bundle-version="3.5.2",
  org.eclipse.core.runtime;bundle-version="3.5.0",
  org.junit;bundle-version="[4.8.1,5.0.0)",
- org.eclipse.jface;bundle-version="3.10.0"
+ org.eclipse.jface;bundle-version="3.10.0",
+ org.hamcrest.core;bundle-version="1.3.0"
```
alexeykazakov: What do you need org.hamcrest.core for?

alexeykazakov:
testPR

usage/tests/org.jboss.tools.usage.test/META-INF/MANIFEST.MF
mbooth101: This bundle makes explicit use of hamcrest APIs. The version of junit in my environment does not re-export hamcrest, so an explicit requirement is needed. This has the added bonus of protecting you from build failures should the version of junit in orbit ever change to be more in line with upstream junit.

alexeykazakov:
Merged to master.

PR #546, opened 2016-12-19
JBDS-4191 remove dependency on...

nickboldt:
JBDS-4191 remove dependency on com.atlassian.connector.eclipse.* since we no longer ship that in Central


fbricon:
+1

PR #553, opened 2017-01-31
JBIDE-23677 enable maven enforcer test to...

nickboldt:
JBIDE-23677 enable maven enforcer test to ensure we set correct version of foundation.core (based on parent pom version)


nickboldt:
Depends on https://github.com/jbosstools/jbosstools-maven-plugins/commit/157388bd6c116fa91c16130d8a164ec7da207e2e being built and deployed.

PR #558, opened 2017-02-28
JBIDE-24014 fix versions in 4.4.x branch -...

nickboldt:
JBIDE-24014 fix versions in 4.4.x branch - bump to 1.3.4-SNAPSHOT in changed plugins/features

### DIFF
--- a/common/features/org.jboss.tools.common.all.feature/pom.xml
+++ b/common/features/org.jboss.tools.common.all.feature/pom.xml
@@ -8,5 +8,6 @@
 	</parent>
 	<groupId>org.jboss.tools.common.features</groupId>
 	<artifactId>org.jboss.tools.common.all.feature</artifactId> 
+	<version>3.8.4-SNAPSHOT</version>
 	<packaging>eclipse-feature</packaging>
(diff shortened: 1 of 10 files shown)


common/features/org.jboss.tools.common.all.feature/pom.xml
```diff
@@ -8,5 +8,6 @@
 	</parent>
 	<groupId>org.jboss.tools.common.features</groupId>
 	<artifactId>org.jboss.tools.common.all.feature</artifactId> 
+	<version>3.8.4-SNAPSHOT</version>
```
dgolovin: This version is not required, it is basically code duplication when version is the same as in parent pom.xml. The same is true for all other duplicated versions below.

nickboldt: Not true. This is how we differentiate a specific plugin from the root pom's version -- the alternative is to upversion everything in the project, which results in way more changes than are needed. Parent pom has a completely different version - 4.4.3.AM1-SNAPSHOT, to be specific.

PR #582, opened 2017-05-26
jbide-24480: Configure sonar for jbosstools-base

bdshadow:


pom.xml
```diff
@@ -16,6 +16,7 @@
 	<properties>
 		<enforceFailOnUIDependencyInCore>true</enforceFailOnUIDependencyInCore>
  		<tycho.scmUrl>scm:git:https://github.com/jbosstools/jbosstools-base.git</tycho.scmUrl>
+ 		<sonar.test.inclusions>**/*Test.*,**/test/**/*</sonar.test.inclusions>
```
jeffmaury: According to https://github.com/jbosstools/jbosstools-build/blob/master/parent/pom.xml#L290 the list is slightly different

bdshadow:
```xml
<include>**/AllTests.class</include>
<include>**/*AllTests*.class</include>
<include>**/*AllBotTests*.class</include>
<include>**/*TestSuite*.class</include>
```
^ don't cover the ones i included into `sonar.test.inclusions`. But i'll add  them too

bdshadow: added

pom.xml
```diff
@@ -40,4 +41,57 @@
 			</snapshots>
 		</repository>
 	</repositories>
+	
+	<profiles>
+		<profile>
+			<id>quality</id>
+			<activation>
+				<activeByDefault>true</activeByDefault>
```
jeffmaury: Why is it active by default ? What about dev running mvn on their laptop ? jacoco is already configured in the parent POM so why do we need it here ?

bdshadow: good catch. Copy-pasted it from jbostools-openshift. Need to fix it in both here and there now. Left after experimenting with sonar. Then forgot to remove

bdshadow: Removed default activation. 
I've tried this build on my laptop. Everything is fine. About the second question we need to be here so that sonar could pick up the defined properties and know where to take generated jacoco files for the coverage reports.

PR #585, opened 2017-06-27
JBIDE-24567 - fix localization for linux security fragment

robstryker:


robstryker:
This seems to have been merged already

PR #586, opened 2017-07-19
Fix sonar issues in abstract trace class

apupier:


jeffmaury:
testPR

jeffmaury:
@apupier can you update it with the DCO header ?

PR #588, opened 2017-07-24
Iterate over the "entrySet" instead of the "keySet"

apupier:
it improves performance

Signed-off-by: Aurelien Pupier <apupier@redhat.com>


jeffmaury:
testPR

PR #599, opened 2017-09-27
JBIDE-25106 - improve robustness for untar

apupier:
when a tar is containing a file and a directory at the same entry name,
just log it and continue the extraction

I choose to log and continue.
Would you prefer to just throw a better error?


apupier:
the root issue was that Jboss Tools doesn't support PaxHeaders, see https://issues.jboss.org/browse/JBIDE-25113

PR #612, opened 2018-03-21
JBIDE-25832 - choose credential composite should update itself when m…

robstryker:
…odel changes;

Clean up strings referencing jboss.org

Signed-off-by: Rob Stryker <rob@oxbeef.net>


robstryker:
testPR

PR #627, opened 2018-06-27
JBIDE-26098 Update stacks.core to use newer artifacts

odockal:
Repair failing tests.

Signed-off-by: Ondrej Dockal <odockal@redhat.com>

### DIFF
--- a/stacks/tests/org.jboss.tools.stacks.core.test/src/org/jboss/tools/stacks/core/test/StacksCoreTest.java
+++ b/stacks/tests/org.jboss.tools.stacks.core.test/src/org/jboss/tools/stacks/core/test/StacksCoreTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013 Red Hat, Inc.
+ * Copyright (c) 2013-2018 Red Hat, Inc.
  * Distributed under license by Red Hat, Inc. All rights reserved.
  * This program is made available under the terms of the
  * Eclipse Public License v1.0 which accompanies this distribution,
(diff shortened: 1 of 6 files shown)


stacks/tests/org.jboss.tools.stacks.core.test/src/org/jboss/tools/stacks/core/test/StacksCoreTest.java
```diff
@@ -10,23 +10,12 @@
  ******************************************************************************/
 package org.jboss.tools.stacks.core.test;
 
-/*************************************************************************************
- * Copyright (c) 2013 Red Hat, Inc. and others.
```
nickboldt: Generally we add dates, not replace them, IIRC. So (c) 2013, 2018 or 2013-2018 is better here. @jeffmaury WDYT?

stacks/tests/pom.xml
```diff
@@ -15,24 +15,6 @@
...
-                        <plugin>
-                                <groupId>org.eclipse.tycho</groupId>
-                                <artifactId>tycho-surefire-plugin</artifactId>
-                                <version>${tychoVersion}</version>
-                                <configuration>
-                                        <dependencies>
-                                                <dependency>
-                                                        <type>eclipse-plugin</type>
-                                                        <artifactId>org.eclipse.equinox.event</artifactId>
-                                                        <version>0.0.0</version>
```
nickboldt: new 4.9.0.AM1 snapshot TP will include org.eclipse.equinox.event_1.4.300.v20180612-1048
old 4.80.1.Final TP had org.eclipse.equinox.event_1.4.200.v20180426-0845.

I don't know how they're different, just calling out that maybe with a new eclipse 4.9-based TP this will work. We could also move to Tycho 1.2 when building.

jeffmaury:
testPR

stacks/tests/pom.xml
odockal: I tried to run with newer TP with dependency on equinox.event. The same problem, ide is not even started -> no test for stacks are being run. Only removal of this dependency seems to be a solution, yet, I am not the author of this plugin, I do not have a clue why was this dependency there.

odockal:
@jeffmaury Versions were bumped up in meantime, so I did a rebase. 
@nickboldt Due to lack of time and capabilities, I gave up to update all the dependencies of httpx components, I just bumped snakeyaml plugin to 1.18 and org.junit to 4.12. Otherwise, I added test that was not run. Finally, I tried to run test locally in maven and bug that was stopping eclipse to run is gone (caused by equinox even plugin in pom.xml dependency) - no idea why.

robstryker:
testPR

robstryker:
This is really old. Closing.

PR #631, opened 2018-09-13
 [JBIDE-26334] improved ux in launcher wizard page

adietish:
Here's how it now looks:
image
@jeffmaury please review


adietish:
@jeffmaury: @robstryker asking me about this plugin and discussing it seems to confirm my feeling that the launcher plugins should be moved to central:
* base is intended to hold common functionalities that are used by all other JBT plugins
* typically common plugins should therefore be in the very end of dependency chains, holding as little dependencies as possible.
Launcher (which depends on m2e et all) therefore seems not to match what base/commons is intended to hold. Central looks like the better place. 
@robstryker @jeffmaury Thoughts?

jeffmaury:
> @jeffmaury: @robstryker (which was/is maintainer of common/foundation) asked me about this plugin and discussed it with me. These insights seem to confirm my feeling that the launcher plugins should be moved to central:
> 
> * base is intended to hold common functionalities that are used by all other JBT plugins
> * base plugins therefore typically are at the very end of dependency chains, holding as little dependencies as possible.
> 
> Launcher (which depends on m2e et all and provides far more than simple library-like functionalities) therefore seems not to match what base/commons was intended to hold. Central looks like the better home.
> 
> @robstryker @jeffmaury Thoughts?

Well I agree even if I think the dependency issue is not a problem anymore since Core Eclipse bundles M2E but I would not do the move before 4.9.0.Final/12.9.0.GA

adietish:
@jeffmaury works for me. 
Dare to review my patch, please?

robstryker:
This should all live in central, and it shouldn't really be added to the common.all feature if we intend to simply move it later. 

Give it it's own feature, add it to the master list of features added to the project, and then later move it to the central repo. 

I personally believe a rename-and-move is just fine even before a Final. Keeping it where it is through a final actually is what's harder to change. Then it becomes a sort of API that the common.all feature contains this launcher, and later it won't.

jeffmaury:
I think we can remove the feature from common.all as I put it there in order to have it installed in update site and installer which turned out to be false.

jeffmaury:
Works for me but depending on the selected booster, the description of the booster can be shrunked:

image

adietish:
@jeffmaury gotcha, text is cut off if the initial size is not large enough to hold larger texts. fixing it.

adietish:
@jeffmaury 
image

jeffmaury:
Still having issues on Win10:
- the mission description is cropped
- the Maven section is hidden

image

adietish:
@jeffmaury I now resize to the required vertical size. Behaves now correctly on your Win10?

jeffmaury:
Still have issue with the Maven section:

image

PR #633, opened 2018-10-31
JBIDE-26461 - Adding org.eclipse.jdt.core.manipulation

odockal:
And must-to-do components version bump-up.


odockal:
Thanks @jeffmaury

adietish:
@odockal just out of curiosity: what makes us need jdt.manipulation?

odockal:
@adietish It is used in manifest.mf files in jbt-base, server, javaee and webservices. To what usage? In javaee.batch it is used to retrieve the editor's line delimiter via stubutility, for example. I did not checked the other repos/components.

PR #639, opened 2019-03-12
provide xjc for java 11 support

jeffmaury:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>


apupier:
having everything in same bundle is fine?
For the exported package, I added everything that was proposed, is it fine?

apupier:
the license for the new embedded library is CDDL/GPL1.1 (can be seen on https://mvnrepository.com/artifact/com.sun.xml.bind/jaxb-xjc )
I think CDDL is fine but I have not found how to verify that

apupier:
the failing test is

https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/jbosstools-base-Pull-Request/254/testReport/junit/org.jboss.tools.foundation.core.test.ecf/URLTransportUtilTest/testLongUrl/

Stacktrace

java.lang.NullPointerException
	at org.jboss.tools.foundation.core.test.ecf.URLTransportUtilTest.testLongUrl(URLTransportUtilTest.java:79)

Standard Output

Responding to request: Hello
Response generated

apupier:
> the license for the new embedded library is CDDL/GPL1.1 (can be seen on https://mvnrepository.com/artifact/com.sun.xml.bind/jaxb-xjc )
> I think CDDL is fine but I have not found how to verify that

I understand that CDDL is fine from https://en.wikipedia.org/wiki/Common_Development_and_Distribution_License
"The resulting compiled software product ("binary") can be licensed and sold under a different license, as long as the source code is still available under CDDL"

apupier:
> I don't get how my PR can influence this test as when launched on java 8, the bundle should be ignored and classes from the JVM should be used

the test is failing on master too https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Devstudio/view/devstudio_master/job/jbosstools-base_master/3595/testReport/ this is not related to this PR

apupier:
> having everything in same bundle is fine?

the common.jaxb bundle will need to export all packages if we want to have a separated one.
(otherwise we will get this kind of error Caused by: java.lang.ClassNotFoundException: com.sun.xml.bind.api.ErrorListener cannot be found by org.myseparated.bundle.for.jaxb-xjc-2.2.6 )

apupier:
> For the exported package, I added everything that was proposed, is it fine?

Based on my previous comment, I think that all packages needs to be exported to avoid any other potential issues

adietish:
In terms of license I deem things to be compatible:
GPL is not compatible with EPL. CDDL on the other hand is, both are weak copyleft licenses. XJC in it's header states that CDDL or GPL are options that the user/contributor/distributor can choose or keep the choice further intact. If we package in EPL plugins, my limited understanding would tell me that we're limiting the licensing to CDDL and thus should maybe specify that.

adietish:
I'd also agree on exporting everything to prevent class not found exceptions. It's imho not to the plugin to limit available classes to the ones that we currently use. Future usage would be limited by such a move.

adietish:
@apupier so that I understand, what's the motivation for xjc in JBT?

apupier:
> @apupier so that I understand, what's the motivation for xjc in JBT?

Fuse Tooling is using XJC for the Data Transformation tooling https://github.com/jbosstools/jbosstools-fuse/blob/ad91dbdcc0611a815f1c4d722fa58f828c285705/transformation/plugins/org.jboss.tools.fuse.transformation.core/src/org/jboss/tools/fuse/transformation/core/model/json/JsonModelGenerator.java#L25

adietish:
@apupier ok, makes sense. One could argue that given that fuse is requiring it, it could be kept within xjc. But yes, having it in JBT/commons can make sense.

adietish:
@apupier the other thing I dont understand yet is how this is related to java11 (title). Can you please explain?

apupier:
> @apupier the other thing I dont understand yet is how this is related to java11 (title). Can you please explain?

in Java 8, xjc is part of the JDK. it has been removed in Java 11

adietish:
build fails because of a failing test. See https://issues.jboss.org/browse/JBIDE-26572. Compilation works, and there's minor risk that this addition would have repercussions to existing code. Given the urgency, we'll merge this now and fix the test right afterwards.
Merging, closing.

PR #640, opened 2019-03-13
JBIDE-26572: org.jboss.tools.foundation.core.test.ecf.URLTransportUti…

jeffmaury:
…lTest.testLongUrl failure


adietish:
@jeffmaury looks like the 404 that we had yesterday, which caused our integration-test to fail, doesn't occur today any more. The test is successful. Nevertheless your patch helps to make the suite self-sufficient and thus more solid. Reviewing.

foundation/tests/org.jboss.tools.foundation.core.test/src/org/jboss/tools/foundation/core/test/ecf/URLTransportUtilTest.java
```diff
@@ -261,5 +234,21 @@ public void testCaching() {
...
+			consumer.accept(server);
+		} finally {
+			try {
+				server.stop();
+			} catch (Exception e) {}
+		}		
+	}
+
+	@FunctionalInterface
+	interface ConsumerWithException {
```
jeffmaury: Do you really need this interface ? Can't you use java.util.function.Consumer ?

adietish: @jeffmaury yes because Consumer does not allow throwing Exceptions which our consumers do via asserts etc. Do you see a better solution?

jeffmaury: assert throws unchecked exceptions so that should not be an issue but I think we have to deal with IOException and other kind of exceptions.

adietish: yes, sorry, misstated, the checked exceptions like IOExceptions are the problem.

PR #641, opened 2019-03-15
Remove spurious executable bits from files

mbooth101:
There are some files in the repo with the execute bit set, but are
clearly not files meant to be executed. You can find such files
with e.g.:

$ find * -type f -executable

This changes the permissions on these file to 644 from 755

Signed-off-by: Mat Booth <mat.booth@redhat.com>


adietish:
@mbooth101 thx for pointing this out.

adietish:
build is green for me. Merging.

PR #642, opened 2019-03-27
JBIDE-26598 - provide utilities to know if targeted VM is running Jigsaw

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>

### DIFF
--- a/common/plugins/org.jboss.tools.common.jdt.debug/src/org/jboss/tools/common/jdt/debug/tools/internal/Tools.java
+++ b/common/plugins/org.jboss.tools.common.jdt.debug/src/org/jboss/tools/common/jdt/debug/tools/internal/Tools.java
@@ -426,6 +426,38 @@ public class Tools implements IPreferenceChangeListener, IToolsConstants, IPrope
 		}
 	}
 
+	/**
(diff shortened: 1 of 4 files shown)


common/plugins/org.jboss.tools.common.jdt.debug/src/org/jboss/tools/common/jdt/debug/tools/internal/Tools.java
```diff
@@ -422,6 +427,43 @@ public void invokeLoadAgent(Object virtualMachine, String path, String options)
...
+			}
+			throw new ToolsCoreException(IStatus.ERROR, message, t);
+		} finally {
+			Thread.currentThread().setContextClassLoader(currentLoader);
+		}
+	}
+
+	private boolean isAgentLoadExceptionWithReturnCode0(Throwable cause) {
+		return (cause.getClass().getName().contains("AgentLoadException") && "0".equals(cause.getMessage())) ||
+				(cause.getClass().equals(IOException.class) && "Non-numeric value found - int expected".equals(cause.getMessage()));
```
apupier: I think that in the second case, we are also ignoring when there are real errors.
But in this case, it won't work anyway so we can try to improve it later and check for the _0_ return code

PR #651, opened 2019-09-27
Allow extraction of tar.xz files

robstryker:
Signed-off-by: Rob Stryker <rob@oxbeef.net>


robstryker:
Test added, will make similar changes in rsp. 
Too bad this isn't a shared codebase... :|

PR #658, opened 2020-02-26
JBIDE-27067 Add .settings folders in JSF 2.3 project templates

jkopriva:
Signed-off-by: Josef Kopriva <jkopriva@redhat.com>


jkopriva:
Without these files it is not possible to copy projects.

PR #684, opened 2021-01-12
Add range for javax.activation

apupier:
a range needs to be added because the version 2.0 is using
jakarta.activation.\*, no more javax.activation.\*

the issue appeared with introduction of javax.activation 2.0.0 in Target Platform


jeffmaury (changes requested):
The whole common arbo needs to be dumped

jeffmaury (approved):
LGTM

sbouchet (approved):
LGTM.

PR #689, opened 2021-03-12
JBIDE-27679: Add support for Red Hat SSO login

jeffmaury:
- Also support any OAuth authorization server


sbouchet (commented):
can connect using the "connect to sso" button, but webpage displayed something related to vscode extension.. otherwise the token is correctly displayed after each action on the button

jeffmaury:
> 
> 
> can connect using the "connect to sso" button, but webpage displayed something related to vscode extension.. otherwise the token is correctly displayed after each action on the button

Yes because that's the name of the client-id that is shared by the equivalent vscode extension

common/plugins/org.jboss.tools.common.oauth.core/plugin.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.5"?>
+<plugin>
+   <extension-point id="authorizationServer" name="Authorization server" schema="schema/authorization.exsd"/>
```
sbouchet, line 4: wrong schema name. should be :
```suggestion
   <extension-point id="authorizationServer" name="Authorization server" schema="schema/authorizationServer.exsd"/>
```

common/tests/org.jboss.tools.common.oauth.core.test/plugin.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sbouchet, line 1: please remove if empty.

common/tests/org.jboss.tools.common.oauth.core.test/pom.xml
```diff
@@ -0,0 +1,29 @@
...
+	<parent>
+		<groupId>org.jboss.tools.common</groupId>
+		<artifactId>tests</artifactId>
+		<version>3.16.0-SNAPSHOT</version>
+	</parent>
+	<groupId>org.jboss.tools.common.tests</groupId>
+	<artifactId>org.jboss.tools.common.oauth.core.test</artifactId>
+	<packaging>eclipse-test-plugin</packaging>
+
+   <build>
```
sbouchet, line 14: remove entierly as JUnit5 single unit test should be ok now.

sbouchet: went too fast. the build fails because no unit test found. please revert and remove only the skip configuration.

sbouchet (approved):

PR #690, opened 2021-03-25
Bump default version to 4.19.0.Final

sbouchet:


jeffmaury (approved):
Do we really needs to bump the version as this should have been done as AM1 start ?

sbouchet:
> Do we really needs to bump the version as this should have been done as AM1 start ?

that's an habit for every releases, like #683 . 
i'll revert locally and see if the version checks fails

jeffmaury (approved):
LGTM

PR #691, opened 2021-03-26
JBIDE-27680: Login/bootstrap Developer Sandbox

jeffmaury:
- Review OAuth framework to return IDToken
- Returned token can be IDToken (default), access token or refresh token
- Add a new method to SwtUtil to deal with top shell and optimum size


sbouchet (approved):
LGTM

PR #701, opened 2021-06-23
JBIDE-27899: Login dialog: renamed 'Closed' button 'Cancel' 

adietish:


common/plugins/org.jboss.tools.common.oauth.ui/src/org/jboss/tools/common/oauth/internal/ui/BrowserBasedLoginDialog.java
```diff
@@ -39,7 +37,7 @@ public BrowserBasedLoginDialog(Shell parentShell, KeycloakDeployment deployment)
 
 	@Override
 	protected void createButtonsForButtonBar(Composite parent) {
-		createButton(parent, IDialogConstants.OK_ID, "Close", true);
+		createButton(parent, IDialogConstants.OK_ID, "Cancel", true);
```
jeffmaury, line 40: I think it's misleading as Cancel semantic is to interrupt current processing which is not what this button does

adietish: @jeffmaury I believe that "Cancel" is exactly what this button does, it does not simply close the dialog, it cancels the process that the user is in: pick "Login" in the context menu to your cluster and proceed until you get "Display Token". But instead of clicking this link, hit "Close". Back in the login form you don't get the token. You actually aborted retrieving the token.

jeffmaury: I think you're confusing the SSO dialog and the OpenShift one. As the button is an OK button, if you click the button while SSO finished the handshake, then there is a small window where user might get the retrieved token when he thought Cancel. So I think we need to change the ID for the button not only the label

adietish: @jeffmaury corrected the constant to `IDialogConstants.CANCEL_ID`

adietish:
@jeffmaury please review

jeffmaury (approved):
LGTM

PR #705, opened 2021-08-11
JBIDE-27965: remove EOL RichFaces

sbouchet:


jeffmaury (dismissed):
LGTM. Maybe bumping the minor would be my choice (as this is not only a fix)

sbouchet:
> LGTM. Maybe bumping the minor would be my choice (as this is not only a fix)

going to do so and finally did not :) I'll update the minor.

jeffmaury (approved):
LGTM

PR #711, opened 2021-09-22
warn user about sec exception to allow non-https

adietish:
This PR introduces a warning that informs Mac users about an exception to the default security policy (non-https connections are forbidden).
image

### DIFF
--- a/foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
+++ b/foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
@@ -10,7 +10,6 @@
  ******************************************************************************/
 package org.jboss.tools.foundation.ui.util;
 
-import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


adietish:
@jeffmaury please review

jeffmaury (commented):
I think that it should be made clear that choosing No will break the current workflow. Also as we know this does not work with CRS, may be a link to an article on tools.jboss.org explaining how to change the .plist would be good.

adietish:
>  may be a link to an article on tools.jboss.org explaining how to change the .plist would be good.

I linked the comment in the issue instead: 
image

jeffmaury:
> 
> 
> > may be a link to an article on tools.jboss.org explaining how to change the .plist would be good.
> 
> I linked the comment in the issue instead:
> image

I can't see those strings reading the code, am I missing something ?

foundation/plugins/org.jboss.tools.foundation.ui/src/org/jboss/tools/foundation/ui/util/BrowserUtility.java
```diff
@@ -372,4 +383,71 @@ public static void allowNonHttpsRedirects() {
...
+				MessageDialog.QUESTION,
+				new String[] { IDialogConstants.YES_LABEL, IDialogConstants.NO_LABEL },
+				0,
+				"Don't ask again",
+				false) {
+
+			@Override
+			protected Control createCustomArea(Composite parent) {
+				Link link = new Link(parent, SWT.None);
+				link.setText("For Red Hat CodeReady Studio manually changing Info.plist within the application bundle is needed. See <a>JBIDE-28007</a>.\n");
```
jeffmaury, line 413: Maybe worth replacing Red Hat CodeReady Studio by Custom IDEs based on Eclipse IDE ?

adietish:
> I can't see those strings reading the code, am I missing something ?
You're right, the state in the main bundle is never read. There are several ways how TSA can be set. So instead of trying to find out whether TSA has been enabled for HTTP I go for me having set it already.

adietish:
@jeffmaury I'm now reading the existing values. Please review.

jeffmaury (approved):
LGTM

PR #713, opened 2021-09-30
JBDS-5032: force depend com.sun.xml.bind 2.2.0

sbouchet:


jeffmaury (commented):
Don't we need a version bump ?

sbouchet:
> Don't we need a version bump ?

yes we should. as there parent did not changed and I did not pushed the stable at the time  it was not falling for baseline comparison.

jeffmaury (approved):
LGTM

PR #730, opened 2022-03-04
JBIDE-28356: fix timeout issue

sbouchet:

### DIFF
--- a/foundation/tests/org.jboss.tools.foundation.core.test/src/org/jboss/tools/foundation/core/test/ecf/URLTransportUtilTest.java
+++ b/foundation/tests/org.jboss.tools.foundation.core.test/src/org/jboss/tools/foundation/core/test/ecf/URLTransportUtilTest.java
@@ -153,19 +153,15 @@ public class URLTransportUtilTest extends TestCase {
 	@Test
 	public void testTimeoutWithoutCancel() throws Exception {
 		executeWithHttpServer(server -> {
-			try {
-				String serverUrl = server.getURI().toURL().toExternalForm();
-				String url = URI.create(serverUrl.substring(0, serverUrl.lastIndexOf(':'))).toURL().toExternalForm()+"/";
-				final IProgressMonitor monitor = new NullProgressMonitor();
-				URLTransportUtility util = new URLTransportUtility();
-				ByteArrayOutputStream os = new ByteArrayOutputStream();
-				long t1 = System.currentTimeMillis();
-				IStatus s = util.download("displayString", url, os, 500, monitor);
-				long t2 = System.currentTimeMillis();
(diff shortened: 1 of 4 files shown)


foundation/tests/org.jboss.tools.foundation.core.test/src/org/jboss/tools/foundation/core/test/ecf/URLTransportUtilTest.java
```diff
@@ -150,16 +151,16 @@ protected Server startFakeHttpServer(AbstractHandler handler, int status, final
 	}
 	
 	@Test
-	@Ignore
 	public void testTimeoutWithoutCancel() throws Exception {
 		executeWithHttpServer(server -> {
 			try {
-				URL url = server.getURI().toURL();
+				String serverUrl = server.getURI().toURL().toExternalForm();
```
jeffmaury, line 157: Not sure to understand why the test needs to be modified ? Are users expected to do the same ?

sbouchet: i found out while debugging that the default URL used before does not trigger the timeout error as expected. 
when the URL is requested, it goes to the mock handler and wait according to the delay, then respond OK. 
so the modification here is just to remove the port from the fake server to simulate a more real timeout because when port is missing, server does not respond to the request.

sbouchet: so users should not modify their logic.

jeffmaury: So I don't understand because the test logic is that the server takes 45s to respond and the request is given 500ms to process thus the timeout

sbouchet: yes, the test should be working that way but it does not anymore. the requested url respond to the request *then* wait for 45s. the timeout does not occurs anymore in that case and the returned staus is OK. As the implementation changed, it here tried to use a know url that will for sure create a real timeout.

sbouchet: so, so what's about his last change? i'm using a third-party service that emulates any kind of URL code (http://httpstat.us/).
using  the http://httpstat.us/408 as request timeout makes it.

jeffmaury (approved):
LGTM

PR #756, opened 2023-05-22
JBIDE-28993 : fix build after tp udpate

sbouchet:


olkornii (approved):
@sbouchet Looks good for me. We have also removed tests for resource navigator in RedDeer.

PR #758, opened 2023-06-05
Common to foundation

robstryker:
This PR has failing tests locally, but hte failing tests already existed locally. Let's see if this builds. 

This PR is copying some classes from common to foundation on the expectation we remove a lot of deps on wtp and such in the future.


robstryker:
This PR is not the cause of the failing check. I made another PR with a simple addition of a plain .txt file and the result is the same. 

https://github.com/jbosstools/jbosstools-base/pull/759

Something in ECF has changed which has made the timeout feature of downloading files via ecf to fail suddenly.

robstryker:
Bypassed rules to push in.

PR #760, opened 2023-06-08
send to segment, not google anayltics (JBIDE-28396)

adietish:
fixes https://issues.redhat.com/browse/JBIDE-28396


usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/internal/telemetry/core/configuration/TelemetryConfiguration.java
```diff
@@ -0,0 +1,148 @@
...
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+import org.jboss.tools.usage.internal.telemetry.core.util.Directories;
+
+public class TelemetryConfiguration extends CompositeConfiguration {
+
+    public static final String KEY_MODE = "com.redhat.devtools.intellij.telemetry.mode";
```
sbouchet, line 21: why is it the same key as Intellij ?

sbouchet (commented):
see above remarks

usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/event/UsageReporter.java
```diff
@@ -306,7 +310,7 @@ protected boolean isPreferencesEnabled() {
 
 	protected EventSender getEventSender() {
 		if(eventSender==null) {
-			eventSender = new GoogleAnalyticsEventSender();
+			eventSender = new TelemetrySender();
```
adietish, line 313: this is the main switch where we stop sending to google analytics but send to segment instead.

adietish:
@sbouchet: thanks for pointing out
> why is it the same key as Intellij
it now saves to a separate file with a separate key:
~/.redhat/org.jboss.tools.usage.internal.telemetry:
```
org.jboss.tools.usage.internal.telemetry.mode=DEBUG
```
The file is automatically created as soon as you agree to tracking of your usage. If it already exists, nothing in it is changed. So if you switch it to DEBUG (from NORMAL), your key is preserved.
In other words enablement is in a first step decided by the preferences (as it was before this PR). Then in a 2nd step segment telemetry decided upon the above file whether tracking information is sent to the NORMAL- or DEBUG-target or not sent at all (DISABLED).

sbouchet:
> @sbouchet: thanks for pointing out
> 
> > why is it the same key as Intellij
> 
> it now saves to a separate file with a separate key: **~/.redhat/org.jboss.tools.usage.internal.telemetry**
> 
> ```ini
> org.jboss.tools.usage.internal.telemetry.mode=DEBUG
> ```
> 
> The file is automatically created as soon as you agree to tracking of your usage. If it already exists, nothing in it is changed. So if you switch it to DEBUG (from NORMAL), your key is preserved. In other words enablement is in a first step decided by the preferences (as it was before this PR). Then in a 2nd step segment telemetry decided upon the above file whether tracking information is sent to the NORMAL- or DEBUG-target or not sent at all (DISABLED).

this is great.
it also means users need to re-enable telemetry if they agreed before, as the key changed? 
this is ok, since we moved out from GA to Segment, it makes sense to re-ask users to agree on telemetry again. 
CC @fbricon for the record.

sbouchet:
> @sbouchet, @fbricon:
> 
> > it also means users need to re-enable telemetry if they agreed before, as the key changed?
> 
> Unfortunately this isn't the case. If you agreed to the dialog ("do you allow us to report usage?") and your decision made it to the preferences, then things stay as they are, you are not asked again, reporting wont happen. This is the case because usage queries the preferences in a first step and only then queries segment/telemetry. If we wanted to force a new ack onto users, we'd have to reset the preferences value. I'd argue that there's no change in data, we only send to a different storage. A given agreement therefore should stay valid. If this wasn't the case we would have had to query people in telemetry again when we switched from woopra to amplitude.
> 
> There's a 2nd shortcoming: the value in the telemetry config file is only loaded once, when queried for the first time. In other words, if you change the value while it is running, the change is not taken into account. It only is when Eclipse is restarted.

Ah thanks for the explanation. i was trying to understand if the actual users that are sending to GA will get a new dialog asking again. It's not, because this dialog does not look into the key, but a preference value.
So all users that ack fortelemetry will send now to segment without a new question.
It's not that i want to ask users about the changes, i'll make a annoucement on the blog about this ( changing telemetry back from GA to Segment ) to make it publicly visible.

sbouchet (dismissed):
LGTM

fbricon:
Since we're collecting exactly the same data, only changing the processing backend, and everything still abides by the Red Hat privacy statement the users agreed on, we don't need to ask again

usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/telemetry/TelemetrySender.java
```diff
@@ -0,0 +1,69 @@
+package org.jboss.tools.usage.telemetry;
```
fbricon, line 1: header

fbricon (changes requested):
a bunch of new files have 2022 or 2022 in their copyright header. If the code was copied from somewhere else, it'd be nice to link to the original code

usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/internal/telemetry/core/service/IDE.java
```diff
@@ -0,0 +1,53 @@
...
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.usage.internal.telemetry.core.service;
+
+import org.jboss.tools.usage.internal.JBossToolsUsageActivator;
+import org.jboss.tools.usage.internal.environment.eclipse.IEclipseUserAgent;
```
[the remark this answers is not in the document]
adietish: I'd basically agree with you but if done consistently this would lead to lots of changes. I'll limit this to a "sane" extent

usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/internal/telemetry/core/util/TimeUtils.java
```diff
@@ -0,0 +1,78 @@
...
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.Date;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class TimeUtils {
```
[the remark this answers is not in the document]
adietish: yes, ex. in ShutdownMessage via static import `TimeUtils.toLocalTime`:
```
       private ShutdownMessage(ServiceFacade service) {
            this(toLocalTime(ApplicationManager.getApplication().getStartTime()), service);
```

usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/internal/telemetry/core/service/segment/IdentifyTraitsPersistence.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Persistency for {@link IdentifyTraits}.
+ */
+public class IdentifyTraitsPersistence {
+
+    private final UsagePluginLogger logger = JBossToolsUsageActivator.getDefault().getLogger();
+
+    public static final IdentifyTraitsPersistence INSTANCE = new IdentifyTraitsPersistence();
+    private static final Path FILE = Directories.RED_HAT.resolve("segment-identify-traits.json");
```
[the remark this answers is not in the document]
adietish: In other words, you want Eclipse and IntelliJ to use different traits? Wouldn't it make sense to know if a user is using Eclipse and IntelliJ?

usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/internal/telemetry/core/service/TelemetryServiceFactory.java
```diff
@@ -0,0 +1,39 @@
...
+import org.jboss.tools.usage.internal.telemetry.core.service.segment.SegmentBroker;
+import org.jboss.tools.usage.internal.telemetry.core.service.segment.SegmentConfiguration;
+
+public class TelemetryServiceFactory {
+
+    private final Environment.Builder builder = new Environment.Builder()
+    		.ide(new IDE.Factory().create().setJavaVersion());
+
+    public TelemetryService create(Plugin plugin) {
+        Environment environment = builder.plugin(plugin).build();
```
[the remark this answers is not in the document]
adietish: I agree, doing.

adietish:
@fbricon thanks for the inputs. Corrected these, please re-review.

sbouchet (dismissed):
LGTM

usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/internal/telemetry/core/service/segment/SegmentBroker.java
```diff
@@ -0,0 +1,292 @@
...
+								runnable.run();
+							} finally {
+								currenThread.setContextClassLoader(backup);
+							}
+						}
+
+					})
+					.build();
+		}
+    }
```
fbricon, line 290: indentation is meh

usage/plugins/org.jboss.tools.usage/src/org/jboss/tools/usage/internal/telemetry/core/service/segment/IdentifyTraits.java
```diff
@@ -0,0 +1,70 @@
...
+ * Contributors:
+ * copied from https://github.com/redhat-developer/intellij-redhat-telemetry 
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.usage.internal.telemetry.core.service.segment;
+
+import java.util.Objects;
+
+/**
+ * Traits that the segment broker sends with am identify event.
```
fbricon, line 17: an indentify event

fbricon:
@adietish feel free to merge once the small nits are addressed

fbricon (approved):

